PR #23, opened 2016-09-19
Fixed removing factors not cascading

### LINKED ISSUE #55: ANOVA UI: Removing factors does not cascade correctly
in this case, i had the fixed factors `Gender` and `Region`, i then assigned these to the plot, then i removed `Gender`. as can be seen, `Gender` remains assigned to `Separate lines`

screen shot 2016-12-02 at 18 01 09

dropmann:
- closes #55

### DIFF
--- a/ui/anova.src.js
+++ b/ui/anova.src.js
@@ -349,6 +349,7 @@ var anovaLayout = LayoutDef.extend({
         context.setValue("modelTerms", currentList);
 
         this.updateContrasts(context, variableList);
+        this.updateFactorDependents(context, variableList, currentList);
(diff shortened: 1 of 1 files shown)


ui/anova.src.js
```diff
@@ -426,6 +427,40 @@ var anovaLayout = LayoutDef.extend({
         context.setValue("contrasts", list3);
     },
 
+    cleanDependantOption: function(context, sourceList, dependantName, dependantIsList) {
```
jonathon-love: spelling, 'dependent'

PR #33, opened 2016-10-11
changed LayoutDef to ui

dropmann:


jonathon-love:
sooo this wasn't quite what i had in mind.

what the ui files call LayoutDef 'internally' doesn't really matter. it's what they export which is key. so i mean changing:

```
module.exports = { LayoutDef : layout, options: options };
```

to

```
module.exports = { ui : layout, options: options };
```

PR #34, opened 2016-10-13
Updated varible target list filtering

dropmann:


ui/ancova.src.js
```diff
@@ -24,7 +24,8 @@ var ancovaLayout = ui.extend({
                     type:"variabletargetlistbox",
                     label: "Dependent Variable",
                     showColumnHeaders: false,
-                    variableFilter: ["continuous"],
+                    suggestedVariableTypes: ["continuous"],
```
jonathon-love: this information should come from the yaml, not the .src.js ...

inst/jamovi/ancova.a.yaml
```diff
@@ -12,19 +12,32 @@ options:
     - name: dependent
       title: Dependent
       type: Variable
+      suggested:
+        - continuous
+      permitted:
+        - nominal
+        - ordinal
```
jonathon-love, line 19: so how does this work? continuous is suggested, but not permitted?

PR #56, opened 2016-12-15
Various updates

dropmann:
- added drop behaviour
- fixed major ANCOVA issues
- improved definition syntax


jonathon-love (commented):
you've deleted `ancova.src.js` ? doesn't every analysis need a .src.js file?

dropmann:
something is weird with that file being deleted because i have the file still in my directory

dropmann:
@jonathon-love no it just looks like i've deleted it but i haven't. I've removed a lot of lines but only added 14.

PR #70, opened 2017-01-17
RM ANOVA: switched to Base64 naming

raviselker:


R/anovarm.b.R
```diff
@@ -10,6 +10,36 @@ anovaRMClass <- R6::R6Class(
...
+                modelFormula <- private$.modelFormula()
+                
+                suppressWarnings({
+                    
+                    result <- try(afex::aov_car(modelFormula, data, type=self$options$ss, factorize = FALSE), silent=TRUE)
+                    
+                }) # suppressWarnings
+                
+                if (isError(result)) {
+                    jmvcore::reject(format('\n{}', extractErrorMessage(result)), code="error")
```
jonathon-love, line 33: if you're feeling the need to add a newline here, would all errors benefit from having a newline character placed before them? should we add it into `reject()`?

raviselker: It was just a temporary thing to distinguish between the jamovi part of the error and the R message. Not necessary to add this to `reject()` (and I will change it in a future commit).

PR #72, opened 2017-01-18
RM ANOVA: added effect sizes

raviselker:
Please forgive me the ugliness of the code (I'm aiming at functionality at the moment, I'll clean up and improve the speed of the code asap)

### DIFF
--- a/R/anovarm.b.R
+++ b/R/anovarm.b.R
@@ -468,6 +468,8 @@ anovaRMClass <- R6::R6Class(
                 
                 suppressWarnings({
                     
+                    table$setStatus('running')
+                    
                     referenceGrid <- lsmeans::lsmeans(result, formula)
                     tukey <- summary(pairs(referenceGrid, adjust='tukey'))
                 
@@ -520,6 +522,7 @@ anovaRMClass <- R6::R6Class(
                     table$setRow(rowNo=i, values=row)
                     private$.checkpoint()
                 }
+                table$setStatus('complete')
             }
         },
         .prepareDescPlots=function(data) {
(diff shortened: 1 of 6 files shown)


R/anovarm.b.R
```diff
@@ -112,10 +113,15 @@ anovaRMClass <- R6::R6Class(
...
-            levene <- self$results$get('assump')$get('eqVar')
+            leveneTable <- self$results$get('assump')$get('eqVar')
             rmVars <- sapply(self$options$rmCells, function(x) return(x$measure))
             for (var in rmVars)
-                levene$addRow(rowKey=var, list(name=var))
+                leveneTable$addRow(rowKey=var, list(name=var))
+            
+            if (length(self$options$bs) == 0) {
+                for (var in rmVars)
+                    leveneTable$addFootnote(rowKey=var, "name", "Because there are no between subjects factors specified this assumption is always met.")
```
jonathon-love: shouldn't begin a sentence with 'Because'. you can use *As* in it's place; `As there are ...`

is it that the assumption is always met, or is it that it simply doesn't apply?

raviselker: It is met because there is only one group. You could also say that it does not apply but that requires more insight from the user I think.

jonathon-love: > It is met because there is only one group

i don't think that's true, but you're right about the alternative requiring more insight from the user, so i'm happy with it as is.

raviselker: There's always equality of variances if there is only one group, so technically the assumption is met.

jonathon-love: yeah, i guess that's true.

raviselker:
This pull request is ready for merging now. The RM ANOVA can go in the next release!

jonathon-love:
w00t!

FYI, chucking a Reliability Analysis commit part way through a big PR makes life difficult for me.

raviselker:
Yeah, I would've rather made a separate pull request for that commit but that's not possible right?

jonathon-love:
yes! you can! you create a separate branch for each PR.

jonathon-love:
you can see (at the moment anyway), i've multiple PRs open on BayesFactor:

https://github.com/richarddmorey/BayesFactor/pulls

jonathon-love:
this will up your git-fu

jonathon-love:
squash-merged!

Issue #74, opened 2017-01-20
RM ANOVA: need a few tweaks


jonathon-love:
sphericity tests accumulate superscripts with changes

<img width="562" alt="screen shot 2017-01-21 at 10 56 59" src="https://cloud.githubusercontent.com/assets/3240247/22169465/60856cae-dfc8-11e6-8e1e-b2ea7176a541.png">

Were you going to remove type 1?

<img width="913" alt="screen shot 2017-01-21 at 10 47 45" src="https://cloud.githubusercontent.com/assets/3240247/22169415/fc36cbbc-dfc7-11e6-80da-88ca46c1565a.png">

<img width="898" alt="screen shot 2017-01-21 at 10 48 27" src="https://cloud.githubusercontent.com/assets/3240247/22169400/e6e41404-dfc7-11e6-9ca9-613108140683.png">

filled circles should probably be placed on top of the error bars rather than underneath (also, what do you think about just dropping the title 'dependent' in RM ANOVA plots?):

<img width="492" alt="screen shot 2017-01-21 at 10 50 29" src="https://cloud.githubusercontent.com/assets/3240247/22169382/c3ce9ab6-dfc7-11e6-8c30-51fab51bdca8.png">

jonathon-love:
so the LH of the posthoc's become sparse (combineBelow), should the RH follow suit? or does it look weird?

<img width="463" alt="screen shot 2017-01-21 at 11 05 55" src="https://cloud.githubusercontent.com/assets/3240247/22169632/a3d6b534-dfc9-11e6-88de-13f8929f0345.png">

jonathon-love:
i wonder if this is why our subscripts appear in a funny way:

https://github.com/jamovi/jmv/blob/master/R/anovarm.b.R#L151

should be:

`p<sub>tukey</sub>`

jonathon-love:
thought about shortening some of these (so they're a bit more memorable, i'd have to look these up to remember which letters were omitted in the abbreviation):

`spherCorrGreenGsser` -> `spherCorrGG`
`spherCorrHuyFdt` -> `spherCorrHF`

these guys could be brought into line with anova.a.yaml (or vice versa)

`effSizeN2` -> `etaSq`
`partEffSizeN2` -> `etaSqP`
`effSizeW2` -> `omegaSq`

the other plot options are prefixed with `descPlots`, does these guys need to fall into line?:

`dispErrBars` -> `descPlotsErrBars`
[...]

PR #75, opened 2017-01-24
RM ANOVA: big tidy up

raviselker:
This should fix the last issues from #74 
Also added documentation to the options

### DIFF
--- a/R/anovarm.b.R
+++ b/R/anovarm.b.R
@@ -40,7 +40,7 @@ anovaRMClass <- R6::R6Class(
                 
                 if (isError(result))
                     jmvcore::reject(extractErrorMessage(result), code='error')
-
+                
                 private$.populateEffectsTables(result)
                 private$.populateSpericityTable(result)
                 private$.populateLeveneTable()
@@ -94,12 +94,12 @@ anovaRMClass <- R6::R6Class(
             }       
         },
         .initSpericityTable=function() {
-            spher <- self$results$get('assump')$get('spher')
+            spherTable <- self$results$get('assump')$get('spherTable')
             for (term in self$options$rmTerms)
-                spher$addRow(rowKey=term, list(name=stringifyTerm(term)))
+                spherTable$addRow(rowKey=term, list(name=stringifyTerm(term)))
         },
         .initLeveneTable=function() {
-            leveneTable <- self$results$get('assump')$get('eqVar')
+            leveneTable <- self$results$get('assump')$get('leveneTable')
             rmVars <- sapply(self$options$rmCells, function(x) return(x$measure))
             for (var in rmVars)
                 leveneTable$addRow(rowKey=var, list(name=var))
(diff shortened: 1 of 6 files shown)


jamovi/anovarm.a.yaml
```diff
@@ -171,70 +212,66 @@ options:
...
+        - none
         - ci
         - se
       default: ci
-      description: >
-        Specifies the Error Bar definition as either "Confidence interval"(default) or "Standard Error"
+      description:
+          R: >
+            \code ('none'), \code{'ci'} (default), or \code{'se'}. Use no error
+            bars, confidence intervals, or standard errors on the plots
```
jonathon-love: `\code ('none')` needs to be with curly braces `\code{'none'}`

this makes is sound like the option prevents *all* error bars:

`Use no error bars, confidence intervals, or standard errors on the plots`

how about:

`Use no error bars, use confidence intervals, or use standard errors on the plots, respectively`

i've also been trying to avoid multiple sentences, so i've been using semi-colons to separate these:

`\code{'none'}, \code{'ci'} (default), or \code{'se'}; use no error bars, use confidence intervals, or use standard errors on the plots, respectively`

but don't mind if you prefer separate sentences.

jamovi/anovarm.a.yaml
```diff
@@ -171,70 +212,66 @@ options:
...
       default:
           - tukey
-      description: >
-        Post Hoc Tests Corrections
+      description:
+          R: >
+            a vector of strings indicating which p-value correction should be
+            used for the post-hoc tests. One (or more) of \code{'tukey'}
+            (default), \code{'scheffe'}, \code{'bonf'} (Bonferroni), or
+            \code{'holm'} can be chosen
```
jonathon-love: the convention i've been using has been to list the possible values first (because if someone's looking something up quickly, they can usually remember how to do it, they just can't remember, say "do i specify it `bonferroni` or `bonf`". how about?:

`one or more of \code{'tukey'} (default), \code{'scheffe'}, \code{'bonf'}, or
\code{'holm'}; use Tukey, Scheffe, Bonferroni and Holm posthoc corrections, respectively`

jamovi/anovarm.a.yaml
```diff
@@ -114,15 +147,21 @@ options:
             title: Huynh-Feldt
       default:
           - none
-      description: >
-        Sphericity corrections
+      description:
+          R: >
+            a vector of strings indicating whether any sphericity corrections
+            are to be computed. One (or more) of \code{'none'}, \code{'GG'}
+            (Greenhouse-Geisser), or \code{`HF`} (Huynh-Feldt) can be chosen
```
jonathon-love: list possible values first?

jamovi/anovarm.a.yaml
```diff
@@ -92,15 +107,33 @@ options:
...
+            title: η²
+          - name: partEta
+            title: partial η²
+          - name: omega
+            title: ω²
+      description:
+          R: >
+            a vector of strings indicating which effect size is to be computed.
+            One (or more) of \code{'eta'}, \code{'partEta'}, or \code{`omega`}
+            can be chosen
```
jonathon-love: list possible values first?

jamovi/anovarm.r.yaml
```diff
@@ -269,10 +269,10 @@ items:
                   title: Huynh-Feldt ε
                   type: number
 
-          - name: eqVar
-            title: Test for Equality of Variances (Levene's)
+          - name: leveneTable
+            title: Equality of variances test (Levene's)
             type: Table
-            visible: (homoTests)
+            visible: (leveneTest)
```
jonathon-love, line 275: we've tended to name things after their function, rather than the test (goal-centred user and all). thoughts?

R/anovarm.b.R
```diff
@@ -351,7 +351,7 @@ anovaRMClass <- R6::R6Class(
                     row[['eta']] <- row[['ss']] / SSt
                     row[['partEta']] <- row[['ss']] / (row[['ss']] + SSr)
                     omega <- (row[['ss']] - (row[['df']] * MSr)) / (SSt + MSr)
-                    row[['omega']] <- if (omega < 0) 0 else omega
+                    row[['omega']] <- if (!is.na(omega) && omega < 0) 0 else omega
```
jonathon-love: good practice to put spaces around these exclamation marks:

```
if ( ! is.na(omega) && omega < 0)
```

it is possible to note notice them otherwise

jamovi/anovarm.r.yaml
raviselker: I named it leveneTest, because R users know that test and calling it something else might confuse them.

jonathon-love: yup, i don't mind either way, but there's a sense in which we're not targeting seasoned R users.

PR #81, opened 2017-02-17
ANOVA: fixed post-hoc tests  + descriptive plots

raviselker:
- Added support for interactions in post-hoc tests
- Added support for separate plots in descriptive plots
- Added option 'none' to the error bar defintion
- Added descriptives plots to non-parametric RM ANOVA

### DIFF
--- a/R/anova.b.R
+++ b/R/anova.b.R
@@ -5,6 +5,7 @@ anovaClass <- R6::R6Class(
     inherit=anovaBase,
     private=list(
         .model=NA,
+        .postHocRows=NA,
         .init=function() {
 
             dependentName <- self$options$dep
@@ -57,24 +58,7 @@ anovaClass <- R6::R6Class(
 
 
             # post hoc
-
-            for (postHocVar in self$options$postHoc) {
-
-                if (length(postHocVar) != 1)
-                    next()
-
-                table <- postHocTables$get(postHocVar)
-
-                postHocVar <- postHocVar[[1]]
-                levels <- base::levels(data[[postHocVar]])
-                combs <- utils::combn(levels, 2)
-                apply(combs, 2, function(comb) {
-                    table$addRow(rowKey=comb, list(
-                        var1=comb[1], var2=comb[2]
-                    ))
-                })
-            }
-
+            private$.initPostHoc()
 
             # descriptives
 
@@ -103,6 +87,9 @@ anovaClass <- R6::R6Class(
                     descTable$addRow(rowKey=row, values=row)
                 }
             }
+            
+            # descriptives plots
+            private$.initDescPlots()
 
         },
         .run=function() {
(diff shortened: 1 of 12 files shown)


R/anova.b.R
```diff
@@ -256,73 +248,182 @@ anovaClass <- R6::R6Class(
...
+                for (i in seq_along(ph))
+                    table$addColumn(name=paste0(ph[i],'2'), title=ph[i], type='text', superTitle='Comparison')
+                
+                table$addColumn(name='md', title='Mean Difference', type='number')
+                table$addColumn(name='se', title='SE', type='number')
+                table$addColumn(name='df', title='df', type='number')
+                table$addColumn(name='t', title='t', type='number')
+                
+                if (pnone)
+                    table$addColumn(name='pnone', title='p', type='number', format='zto,pvalue')
```
jonathon-love: so i think you're better off *always* adding these, and toggling their visibility with the option.

calculating these additionally will add maybe 10ms to the analysis, but means that when someone toggles this option, a round trip to R won't be necessary (which will be super-quick once i implement that properly).

R/anova.h.R
```diff
@@ -135,6 +134,7 @@ anovaOptions <- R6::R6Class(
                 "plotError",
                 plotError,
                 options=list(
+                    "none",
                     "ci",
                     "se"),
                 default="ci")
```
jonathon-love, line 140: cole reckons we should have 'none' as default. thoughts?

R/anovarmnp.b.R
```diff
@@ -18,7 +18,7 @@ anovaRMNPClass <- R6::R6Class(
             name <- measureNames[[i]]
             mat[,i] <- jmvcore::toNumeric(data[[name]])
         }
-
+
```
jonathon-love: what's this!

raviselker: What do you mean?

jonathon-love: so this empty line, has been replaced with an empty line with whitespace in it. we try and avoid whitespace at the end of a line (or an empty line). in fact, the RStudio project file should have prevented this.

R/anova.h.R
raviselker: Isn't it better to have some kind of correction as default? Using normal p-values is pretty bad for post-hoc tests (actually using any type of inference on exploratory analyses is pretty silly).

R/anova.b.R
raviselker: So what do you mean with "toggling their visibility with the option"?

R/anova.h.R
jonathon-love: for the plot, not p-values

raviselker: hmm, apparently I wasn't really awake when I wrote the previous reply. I think a CI is pretty nice right, because it shows immediately which effect you can take seriously and which not.

R/anova.b.R
jonathon-love: using the data-binding. always calling `addColumn()`, but specifying visible as `(postHocCor:none)`, like in ... anova:

https://github.com/jamovi/jmv/blob/master/jamovi/anova.r.yaml#L170

(obviously i've specified it here in the yaml, but you can specify it in the .b.R too)

raviselker: ok, got it

R/anova.h.R
jonathon-love: i'm easy. i don't mind either way.

raviselker: So let's leave the default on "ci" for now

jonathon-love:
merged!

https://github.com/jamovi/jmv/commit/4b57ae55e7af070a5cde390cd5edd5740512f27b

PR #82, opened 2017-02-20
Updated to enableOn functionality

dropmann:


jamovi/anova.u.yaml
```diff
@@ -240,15 +240,16 @@ children:
             optionId: plotError
             checkedValue: ci
             label: Confidence interval
-            events:
-              change: './anova.events::onChange_errBarDef_ci'
+            #events:
+            #  change: './anova.events::onChange_errBarDef_ci'
```
jonathon-love: did you want to tidy this guy up a bit more

jamovi/anova.u.yaml
```diff
@@ -240,15 +240,16 @@ children:
...
             children:
               - type: TextBox
                 name: ciWidth
                 label: Interval
                 suffix: '%'
                 format: number
                 inputPattern: '[0-9]+'
+                enableOn: (errBarDef_ci)
```
jonathon-love, line 250: ah! we don't even need the colon notation!

jonathon-love (commented):
This is a great improvement to our quality of life!

less 168 lines of code, plus 21!

jamovi/anova.u.yaml
dropmann: It depends on what controls are being used but if Lists are used as radio buttons then the radio button outputs a boolean, which is helpful.

jonathon-love: ah, OK, so we still may need the colon for the dropdown lists (if we decided that was necessary). but they'll map onto NMXLists nicely too i expect.

dropmann: yep, because NMXLists are currently displayed as individual checkboxes it works fine. If we make a different control to display them we might need the colon notation.

PR #86, opened 2017-03-31
Added EFA (plus added some things in PCA)

raviselker:


raviselker:
Hmm, it's pretty weird that it doesn't pass my unit test (of course I tested it locally and it passes all tests)

PR #87, opened 2017-04-03
2.0 u.yaml Support

dropmann:


R/bancova.b.R
```diff
@@ -0,0 +1,17 @@
+
+# This file is a generated template, your changes will not be overwritten
```
jonathon-love: none of these banova, bancova, etc. analyses should be in there any more. i assume you've got old .a.yaml files kicking around.

PR #97, opened 2017-04-26
Improvements to dropping into embedded lists

dropmann:


dropmann:
@jonathon-love This failing doesn't look like it is not my fault.

dropmann:
@jonathon-love This failing doesn't look like it is not my fault.

PR #103, opened 2017-05-08
Prepared plots for themes

raviselker:


R/descriptives.b.R
```diff
@@ -123,21 +123,20 @@ descriptivesClass <- R6::R6Class(
...
                 geom_bar(stat="identity") +
                 labs(list(x=NULL, y=ylab)) +
-                theme(
-                    text=element_text(size=16, colour='#333333'),
-                    plot.background=element_rect(fill='transparent', color=NA),
-                    panel.background=element_rect(fill='#E8E8E8'),
-                    axis.title.y=element_text(margin=margin(0,10,0,0)),
-                    axis.text.y=axis.text.y)
+                theme +
+                theme(axis.text.y=axis.text.y,
```
jonathon-love, line 138: can and does this work? i can't see how it can. how does it know which theme is which?

raviselker: Why wouldn't it work?

jonathon-love: there two things called `theme`, one is a theme object, the other is a function. how does it know which is which?

raviselker: they exist in different environments I guess

jonathon-love: so i guess you've tried it, and it works, but i'm still pretty surprised.

raviselker: Yup, it does work

PR #104, opened 2017-05-14
Fix to ANCOVA posthoc supplier list update issue

dropmann:


jonathon-love:
Fixes jamovi/jamovi#230

PR #113, opened 2017-06-06
CFA: only execute semPlot when pathDiagram is TRUE

raviselker:


jonathon-love:
you could remove the 'not yet implemented' in the documentation too

PR #115, opened 2017-06-19
Imported model builders

dropmann:
- include interactions transfer action
- fixed model builder issues

### DIFF
--- a/jamovi/anova.u.yaml
+++ b/jamovi/anova.u.yaml
@@ -60,6 +60,7 @@ children:
         format: term
         children:
           - type: TargetLayoutBox
+            transferAction: interactions
             children:
               - type: ListBox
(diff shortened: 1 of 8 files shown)


jamovi/anova.u.yaml
```diff
@@ -60,6 +60,7 @@ children:
         format: term
         children:
           - type: TargetLayoutBox
+            transferAction: interactions
```
jonathon-love, line 63: sooo, assuming this is the existing button in the model builder, adding a single item (which i assume you can do with this button), isn't an interaction but a 'main-effect'. if you're adding all combinations of interactions and main-effects, 'full-factorial' might be a better name.

dropmann:
no this refers to the full two button combo exactly like in jasp

PR #133, opened 2017-12-05
Fix SE difference for Student's t-test

borisma:
SE difference for Student's t-test was based on SE difference for Welch's t-test. Added definition for SE difference for Student's t-test based on the pooled variances (`sediffSTUD`). Thereby, replacing `pooledSD` by `pooledVAR` for clarity and changed definition of Cohen's d accordingly.

### DIFF
--- a/R/ttestis.b.R
+++ b/R/ttestis.b.R
@@ -68,10 +68,11 @@ ttestISClass <- R6::R6Class(
                 se <- sqrt(v/n)
                 sd <- sqrt(v)
 
-                sediff <- tryNaN(sqrt((v[1]/n[1])+(v[2]/n[2])))
-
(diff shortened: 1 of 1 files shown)


raviselker (approved):
Good catch @borisma. The changes look good to me, so they can be merged.

raviselker:
Although the t-test unit test result needs to be changed accordingly. But I can do that after you merged this pull request @jonathon-love

borisma:
Thanks! But the results are not affected since the Student t-test procedure automatically uses the pooled se difference. Right?

Best, Boris

Am 06.12.2017 um 15:03 schrieb Ravi Selker <notifications@github.com<mailto:notifications@github.com>>:


Although the t-test unit test result needs to be changed accordingly. But I can do that after you merged this pull request @jonathon-love<https://github.com/jonathon-love>

raviselker:
Yes, the only thing that should be affected is the pooled SE (the test statistic already used the correct one).

I'm a bit puzzled though, why it doesn't pass the unit test...

PR #136, opened 2017-12-21
RM ANOVA: fixes testthat issue with afex

raviselker:


jonathon-love:
do you need to add emmeans to imports?

raviselker:
I thought you already replaces lsmeans with emmeans at some point...

jonathon-love:
the unit tests are failing because it requires the newer (not released yet) version of afex?

raviselker:
Yes, I assume so

PR #143, opened 2018-02-02
logRegBin: small fixes

raviselker:


jonathon-love:
haha, the challenge is *always* not to have to do a small fix commit after doing a big commit.

PR #146, opened 2018-02-05
Removed legacy code from logregbin

dropmann:


jonathon-love:
vestigal would be a better word here than legacy :)

PR #152, opened 2018-02-28
linReg: aligns analysis with logRegBin

raviselker:
Adds:
- simple model comparison
- possibility to define interactions
- possibility to add factors
- estimated marginal means

### DIFF
--- a/R/linreg.b.R
+++ b/R/linreg.b.R
@@ -4,27 +4,24 @@ linRegClass <- R6::R6Class(
     inherit = linRegBase,
     private = list(
         #### Member variables ----
-        modelSelected = NULL,
-        modelTerms = NULL,
+        terms = NULL,
+        coefTerms = list(),
+        emMeans = list(),
 
         #### Init + run functions ----
         .init = function() {
 
-            if (self$options$modelSelected < 0 || (self$options$modelSelected + 1) > length(self$options$blocks))
-                private$modelSelected <- length(self$options$blocks)
-            else
-                private$modelSelected <- self$options$modelSelected + 1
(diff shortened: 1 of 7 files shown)


R/linreg.b.R
```diff
@@ -1,30 +1,27 @@
 
-linRegClass <- R6::R6Class(
+linRegClass <- if (requireNamespace('jmvcore')) R6::R6Class(
```
jonathon-love: you can get rid of this requireNamespace thing ... no-one will or can use jmv without jmvcore

R/linreg.b.R
```diff
@@ -259,34 +389,93 @@ linRegClass <- R6::R6Class(
                 row[["df2"]] <- r[i,1]
                 row[["p"]] <- stats::pf(row[["f"]], row[["df1"]], row[["df2"]], lower.tail=FALSE)
 
-                BF <- modelsBF[[i+1]]/modelsBF[[i]]
+                # BF <- modelsBF[[i+1]]/modelsBF[[i]]
 
-                row[["bf"]] <- exp(BF@bayesFactor$bf)
-                row[["err"]] <- BF@bayesFactor$error
+                # row[["bf"]] <- exp(BF@bayesFactor$bf)
+                # row[["err"]] <- BF@bayesFactor$error
```
jonathon-love, line 396: should these be tidied up, or are you planning on revisiting?

raviselker: Planning on revisiting (I now got rid of the BF's because they didn't work properly + not sure whether it's appropriate to have them in here)

PR #180, opened 2018-07-20
Added shapiro-wilk to descriptives

jonathon-love:


raviselker (commented):
I wouldn't call the shapiro-wilk test "norm" as there are other normality tests as well. I would just put it under a "Normality tests" heading in the UI.

jonathon-love:
so my plan was to have all normality tests being turned on by this (one) option ...

raviselker:
Hmm, don't you think it's clearer if the option matches the result?

jonathon-love:
yup, but i'm trying to prevent the options from becoming cluttered.

raviselker (approved):

PR #187, opened 2018-08-03
IS T-Test: Added Q-Q plots

jonathon-love:


jonathon-love:
while i've got you here, i'm having doubts about whether this line is correct in descriptives:

https://github.com/jamovi/jmv/blob/master/R/descriptives.b.R#L1020

shouldn't it just be `yAxis <- 45 + 11`?

raviselker:
So about the QQ plot, I think you should use the residuals for this instead (you always test the normality of residuals in general linear models). 

About the other issue, I think it's correct as is. If the level names are longer they're gonna take more space next to y axis.

jonathon-love:
> So about the QQ plot, I think you should use the residuals for this instead (you always test the normality of residuals in general linear models).

won't scale take care of that? the residuals are just subtracting the means, and scale effectively does that?

> About the other issue, I think it's correct as is. If the level names are longer they're gonna take more space next to y axis.

but the level names don't appear beside the y-axis in Q-Q plots:

untitled

raviselker:
So you don't have to separate them in different groups. The overall residuals have to be normally distributed. 

Oh yeah, I thought I was looking at the code for the ridgeplots. For the QQ plots you can get rid of that.

jonathon-love:
so at UoN they want students to look at the groups separately. there's this idea among the stats folk that two non-normal groups might, when viewed together, look normal.

raviselker:
So the assumption is about the normality of the residuals not about the normality of the response variable in different groups. See for instance https://www.theanalysisfactor.com/assumptions-about-residuals/

raviselker (approved):

PR #201, opened 2018-09-16
Added one-way anova + fix lavaan issue

raviselker:


R/anovaonew.b.R
```diff
@@ -0,0 +1,429 @@
...
+        },
+        .prepareQQPlot = function(results) {
+
+            plots <- self$results$plots
+            group <- self$options$group
+
+            for (dep in self$options$deps) {
+
+                image <- plots$get(key=dep)$qq
+                image$setState(results[[dep]]$residuals)
```
jonathon-love: storing one point per row in the data set doesn't scale. if a data set has millions of rows, then that's megabytes of data which is being serialized, and deserialized (slow). so calculate the residuals *in* the plot function itself, like here:

https://github.com/jamovi/jmv/blob/master/R/ttestis.b.R#L501

jamovi/anovaonew.a.yaml
```diff
@@ -0,0 +1,176 @@
...
+      title: Welch's
+      type: Bool
+      default: true
+      description:
+          R: >
+            `TRUE` (default) or `FALSE`, perform Welch's one-way ANOVA which does
+            not assume equal variances
+
+    - name: fishers
+      title: Fisher's
```
jonathon-love: Is there space for `Fisher's (assume equal variances)`? or `Assume equal variances (Fisher's)`?

We try and refer to things by what they're used for (i.e. in our menus)

jonathon-love (changes requested):
Could you separate the lavaan and one-way anova into two commits?

And add references from the commits to the respective issues?

jonathon-love:
oh wait, i see they are separate, and i see you have referenced the issues appropriately.

R/anovaonew.b.R
raviselker: That means that I would also have to do the data cleaning, and fitting the model again in the plot function...

jamovi/anovaonew.a.yaml
raviselker: How about this:
screenshot from 2018-09-16 10-57-51

raviselker: But then it makes more sense to let them be radio buttons :confused:

raviselker:
Gotta pay attention @jonathon-love

R/anovaonew.b.R
jonathon-love: yup.. which is probably all cheaper than disk io - even for small data sets.

you can cache the fitted model (assign it to something `private$.model`) in `.run` (but your plotting code shouldn't assume that `.run` has been called. so you probably want a `.initModel()` function that you call from both `.run()` and your plot function)

jonathon-love: yup.. which is probably all cheaper than disk io - even for small data sets.

you can cache the fitted model (assign it to something `private$.model`) in `.run` (but your plotting code shouldn't assume that `.run` has been called. so you probably want a `.initModel()` function that you call from both `.run()` and your plot function)

jamovi/anovaonew.a.yaml
jonathon-love: Always try and avoid 'yes', 'no' questions.

Something more like:

```
    Variances
        [  ] Assume equal (Fisher's)
        [  ] Don't assume (Welch's)
```

jonathon-love:
oh well, i was looking on my phone, and it certainly looks like all commits are one there.

jonathon-love:
i cherry-picked the CFA and t-test commits.

R/anovaonew.b.R
raviselker: But the fitted model contains all the residuals. So with a big data set, this would also be a big object.

jamovi/anovaonew.a.yaml
raviselker: Don't assume variances?

jonathon-love: Don't assume equal?

R/anovaonew.b.R
jonathon-love: yup, but the private members of an analysis object are not serialized.

jonathon-love: what you assign to the `state` gets serialised and deserialised each time the analysis runs (or when a plot needs to be generated). so we want to keep this smallish.

raviselker: k, so I can just store the residuals into `private$.residuals`?

jonathon-love: yup, but if a plot is requested after the analysis has been run, `.run()` may not be run (i.e. when a person copies a plot to the clipboard), so your plotting code needs to be able to generate residuals if they're not there.

jonathon-love:
the `clearWith` properties of plots look like they need some attention... they refresh when unrelated options change.

raviselker:
Which options are you talking about? Everything seems to work fine over here.

jonathon-love:
ah yup, the changes i made to the `key`s has broken it.

jonathon-love:
merged

PR #231, opened 2019-10-08
ANCOVA: Improved handling of unused levels

jonathon-love:
(you'll also notice i've moved the .checkpoint() call, which wasn't quite in the right place)


raviselker (commented):
It would be nice if you could add a test case for this situation :)

PR #233, opened 2019-11-02
Cont tables: added more totals

jonathon-love:
Closes jamovi/jamovi#779

### DIFF
--- a/R/conttables.b.R
+++ b/R/conttables.b.R
@@ -102,6 +102,37 @@ contTablesClass <- R6::R6Class(
                     type='integer')
             }
 
+            if (self$options$exp) {
+                freqs$addColumn(
+                    name='.total[exp]',
+                    title='Total',
+                    type='number')
+            }
+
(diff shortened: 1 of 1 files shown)


raviselker (changes requested):
I find the `% within $var1$` a lot more confusing then `% within row`. Took me quite a while to understand what these percentages mean with the current way of phrasing.

Screenshot from 2019-11-07 13-13-08

jonathon-love:
so the `% within variableName` is the way that spss does it ... i feel like it's clearer ... 1-10 how strongly are you opposed to it?

raviselker:
I'm probably a 7.5. To me, `% within $level$` (e.g., `% within OJ`) would make more sense...

jonathon-love:
`% within $level` might be problematic in the scenario (which i don't think is too uncommon), where you have `yes`/`no` as the levels in both the row and column variable.

raviselker (approved):

PR #234, opened 2019-11-03
ANCOVA: added cohen's d to posthocs

jonathon-love:


raviselker (changes requested):
- "Effect-size" should be "Effect Size"
- Use the same distance between two groups as in the estimated marginal means (so more space between the correction and effect size groups)
Screenshot from 2019-11-07 13-04-05

raviselker (approved):

raviselker (approved):

PR #244, opened 2020-03-03
AN(C)OVA: Added whole model test

jonathon-love:


jamovi/ancova.a.yaml
```diff
@@ -131,6 +131,16 @@ options:
             one or more of `'eta'`, `'partEta'`, or `'omega'`;
             use η², partial η², and ω² effect sizes, respectively
 
+    - name: modelTest
+      title: Model Test
```
raviselker: Don't capitalize `Test`

jamovi/ancova.u.yaml
```diff
@@ -34,7 +34,15 @@ children:
               change: './ancova.events::onChange_covariates'
   - type: LayoutBox
     margin: large
+    style: inline
     children:
+      - type: Label
+        label: Overall Model
```
raviselker: Maybe use `Model Fit` as the label, and `Overall model test` as the option

jamovi/ancova.u.yaml
```diff
@@ -34,7 +34,15 @@ children:
...
+        margin: large
+        style: list-inline
+        children:
+          - name: modelTest
+            type: CheckBox
```
raviselker: The model test en effect size options are very close together so I would use `stretchFactor` to make it look a bit nicer

R/ancova.b.R
```diff
@@ -371,6 +385,22 @@ ancovaClass <- R6::R6Class(
                         F <- ''
                     if ( ! is.finite(p))
                         p <- ''
+                } else if (index == 0) {
+
+                    # can't use default summary() dispatch here
+                    # aov takes precedence
```
raviselker: I'm not even sure anymore why we use `aov` here instead of `lm`. It has no advantages. It stores way less information, it just gives you a different view to the results. So maybe store the model as an `lm` object instead (make sure you test wether it still works after changing it though)

raviselker (commented):
Just a few little touch ups

jamovi/anova.u.yaml
```diff
@@ -22,8 +22,23 @@ children:
   - type: LayoutBox
     margin: large
```
raviselker, line 23: You need to add an `stretchFactor: 1` to the LayoutBox otherwise the children will be on top of each other

jamovi/ancova.u.yaml
```diff
@@ -34,9 +34,25 @@ children:
               change: './ancova.events::onChange_covariates'
   - type: LayoutBox
     margin: large
```
raviselker, line 36: You need to add an `stretchFactor: 1` to the LayoutBox otherwise the children will be on top of each other

raviselker (changes requested):
Just two minor things with the layout. After you fix those you can merge it in.

PR #254, opened 2020-05-07
Changed default for minimal eigenvalue to 1

sjentsch:
1 is Kaiser's criterium (and even though it is discussed, 0 is rather less appropriate as default)


jonathon-love:
actually, can you run `jmvtools::prepare()` (i think it is), and commit the updated .R.h file too.

PR #255, opened 2020-05-08
Update efa.h.R

sjentsch:
Adjusted default for minEigen


jonathon-love:
so you could have just added this commit to the branch of the other PR you know :)

PR #263, opened 2020-07-14
Added the appropriate effect sizes for different "t-test" statistics

raviselker:
closes jamovi/jamovi#873


jonathon-love (commented):
i think i prefer this, but i won't fight you if you want to keep it the way you've done it.

Screen Shot 2020-07-15 at 10 40 26

raviselker:
Hmm, yeah let me think about that one

PR #264, opened 2020-07-28
AN(C)OVA: fixed post-hoc effect size

raviselker:
- Use the emmeans effect size function instead of "manually" calculating
- Added confidence interval for the effect size
- Added note in table notifying the user that post-hoc tests are based on estimated marginal means

Fixes jamovi/jamovi#914 
Fixes jamovi/jamovi#915


jonathon-love:
merged

PR #270, opened 2020-10-13
added one-sided hypothesis tests for conttables

victor-moreno:
In this new version I have copied code from ttestis to add the option to select one-sided hypothesis for the Fisher and z tests.

Also, I have removed the hybrid method for the >2x2 Fisher test, because in an example used for testing, the p-value was smaller for hybrid than both the chi2 and the exact, which is not reasonable. The hybrid method is not coded in R, so I cannot easily check what exactly does.

### DIFF
--- a/R/conttables.b.R
+++ b/R/conttables.b.R
@@ -199,6 +199,8 @@ contTablesClass <- R6::R6Class(
             }
 
             ciText <- paste0(self$options$ciWidth, '% Confidence Intervals')
+            odds$getColumn('cil[dp]')$setSuperTitle(ciText)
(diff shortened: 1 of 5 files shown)


victor-moreno:
You can test the changes here:
https://jamovi.snpstats.net
There are 2 test datasets with 2x2 and 2x3 tables

jonathon-love:
hi victor,

now that i've merged your first PR, it's brought up some conflicts with this PR. would you be able to resolve them?

with thanks

victor-moreno:
I will close this pull request and open a new one to avoid conflicts

PR #271, opened 2020-10-15
Added one-sided hypothesys to Fisher and DP

victor-moreno:
Changes in this version:
- One-sided hypothesis can be selected for Fisher's test and differences in proportions. This applies only to 2x2 tables and have no effect on larger tables
- The Fisher exact test for rxc tables always used the exact method, not the hybrid as before, because I have seen inconsistencies whe testing the hybrid

### DIFF
--- a/R/conttables.b.R
+++ b/R/conttables.b.R
@@ -405,7 +405,7 @@ contTablesClass <- R6::R6Class(
                         `value[likeRat]`=NaN,
                         `df[likeRat]`='',
                         `p[likeRat]`='',
-                        `value[fisher]`=NaN,
+                        `value[fisher]`='',
                         `p[fisher]`='',
                         `value[N]`=n)
                 } else {
--- a/R/conttables.b.R
+++ b/R/conttables.b.R
@@ -245,6 +245,35 @@ contTablesClass <- R6::R6Class(
             freqRowNo <- 1
             othRowNo <- 1
 
+            # get group names according to compare
+            groups <- NULL
+            variable <- NULL
+            if (self$options$compare == "rows") {
+                if (!is.null(rowVarName)) {
+                    variable <- rowVarName
+                    groups <- base::levels(data[[rowVarName]])
+                } else {
+                    groups <- c('Group 1', 'Group 2')
+                }
+            } else { # compare columns
+                if (!is.null(colVarName)) {
+                    variable <- colVarName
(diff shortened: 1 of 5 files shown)


jonathon-love:
oh, sorry victor, i missed this.

jonathon-love:
one thing i've noticed is that `fish$estimate` is always NULL :

https://github.com/jamovi/jmv/blob/master/R/conttables.b.R#L389

which makes me think `$estimate` isn't quite right. could you take a look at this too?

with thanks

victor-moreno:
Hi Jonathon, 
$estimate was generated by stats::fisher.test()
In a 2x2 table it is the odds ratio, but in a larger table it is always NULL

The exact test doesn't have a statistic, directly provides a p-value. I have changed the code to only show the p-value for the exact test (Value always blank) to avoid misunderstandings with the estimate as Value

Victor

jonathon-love:
thanks victor,

i'll wait for @raviselker to cast his eye over this.

cheers

jonathon

jonathon-love:
someone had a comment here, which might be relevant:

https://forum.jamovi.org/viewtopic.php?f=2&t=1474&sid=5e308352f898d15d57a3dacd0908fed4#p5206

(wrt the issue the thread was originally about, the chi symbol in the table notes was causing issues under windows, but i pushed a fix to jmvcore to address it).

jonathon

victor-moreno:
I have moved the z test to the last and added a 2 in the title: "z test difference in 2 proportions". I don't think it should go to the 2x2 section because the design is the left column for "Tests" and the right column for "Comparative measures 2x2". Before the Fisher test was also only for 2x2 tables. When someone tries a z test on a table larger than 2x2, a footnote warns about the NaN result.

raviselker:
Looks good to me

jonathon-love:
OK, i've merged this (and made some changes on the way, so can you check that my mods are OK?). i flattened it down to 3 commits:

https://github.com/jamovi/jmv/commits/master

so this call:

```
fish <- try(stats::fisher.test(mat, ...
```

will either return a `htest` or a `try-error` ... i can't see that it would ever return `NULL`, so the check:

```
if (is.null(fish)) {
```

didn't work (nor would it ever have worked ... so i'm not quite sure how it got in there).

jonathon

victor-moreno:
Thanks for that Jonathon. Before it worked as expected, at least with the tests I did, but probably by chance

PR #272, opened 2020-11-03
PCA: Fixed bug jamovi/jamovi#960

raviselker:
* It was possible that the parallel analayses returned 0 factors. Now it will always have at least 1 factor


jonathon-love:
merged

PR #275, opened 2020-11-07
Percentiles added to descriptives

victor-moreno:
Currently you can get quantile up to 10 groups in descriptives. I needed the 5th and 95th percentiles, so I propose a new option to generate a comma-delimited list of arbitrary percentiles.
The TextBox is small and I don't know if it could be made larger. 
I have set as default 0.05,0.95 to indicate that any percentile list can be used.

### DIFF
--- a/R/descriptives.b.R
+++ b/R/descriptives.b.R
@@ -1037,6 +1037,7 @@ descriptivesClass <- R6::R6Class(
                 pcValues <- as.numeric(unlist(strsplit(self$options$pcValues,",")))
                 pcValues[pcValues < 0 | pcValues > 1] <- NA 
                 pcValues <- pcValues[!is.na(pcValues)]
+                pcValues <- pcValues[ ! (pcValues %in% pcEq) ]
                 npcValues <- length(pcValues)
 
                 if (npcValues > 0){
--- a/R/descriptives.b.R
+++ b/R/descriptives.b.R
@@ -1032,7 +1032,7 @@ descriptivesClass <- R6::R6Class(
             pcValues <- pcValues[ ! (pcValues %in% pcEq) ]
 
             return(pcValues)
-        }
+        },
         .addQuantiles = function() {
 
             if ( self$options$pcEqGr ) {
--- a/R/descriptives.b.R
+++ b/R/descriptives.b.R
@@ -1104,8 +1104,9 @@ descriptivesClass <- R6::R6Class(
                 npcValues <- length(pcValues)
 
                 if (self$options$pcVal && npcValues > 0) {
+                    quants <- as.numeric(quantile(column, pcValues))
                     for (i in 1:npcValues)
-                        stats[[paste0('perc', i)]] <- pcValues[i]
+                        stats[[paste0('perc', i)]] <- quants[i]
                 }
 
             } else if (jmvcore::canBeNumeric(column)) {
--- a/R/descriptives.b.R
+++ b/R/descriptives.b.R
@@ -1038,12 +1038,12 @@ descriptivesClass <- R6::R6Class(
             pcValues <- pcValues[!is.na(pcValues)]
             npcValues <- length(pcValues)
 
-            if (!self$options$pcVal | npcValues == 0)
+            if (! self$options$pcVal || npcValues == 0)
                 return()       
 
             colArgs <- private$colArgs
 
-            private$colArgs$name <- c(colArgs$name, paste0('quant', 1:npcValues))
+            private$colArgs$name <- c(colArgs$name, paste0('perc', 1:npcValues))
             private$colArgs$title <- c(colArgs$title, paste0(round(pcValues * 100, 2), 'th percentile'))
             private$colArgs$type <- c(colArgs$type, rep('number', npcValues))
             private$colArgs$visible <- c(colArgs$visible, rep("(pcValues)", npcValues))
@@ -1098,6 +1098,16 @@ descriptivesClass <- R6::R6Class(
                         stats[[paste0('quant', i)]] <- quants[i]
                 }
 
+                pcValues <- as.numeric(unlist(strsplit(self$options$pcValues,",")))
+                pcValues[pcValues < 0 | pcValues > 1] <- NA 
+                pcValues <- pcValues[!is.na(pcValues)]
+                npcValues <- length(pcValues)
+
+                if (self$options$pcVal && npcValues > 0) {
+                    for (i in 1:npcValues)
+                        stats[[paste0('perc', i)]] <- pcValues[i]
+                }
+
             } else if (jmvcore::canBeNumeric(column)) {
 
                 l <- list(mean=NaN, median=NaN, mode=NaN, sum=NaN, sd=NaN, variance=NaN,
@@ -1108,6 +1118,15 @@ descriptivesClass <- R6::R6Class(
                 if ( ! (self$options$quart && pcNEqGr == 4)) {
                     for (i in 1:(pcNEqGr-1))
                         l[[paste0('quant', i)]] <- NaN
+                }
+
+                pcValues <- as.numeric(unlist(strsplit(self$options$pcValues,",")))
+                pcValues[pcValues < 0 | pcValues > 1] <- NA 
+                pcValues <- pcValues[!is.na(pcValues)]
+                npcValues <- length(pcValues)
+                if (! self$options$pcVal || npcValues == 0) {
+                    for (i in 1:npcValues)
+                        l[[paste0('perc', i)]] <- NaN
                 }
 
                 stats <- append(stats, l)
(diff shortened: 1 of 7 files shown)


jonathon-love:
i wonder if we shouldn't default this to 0.25, 0.5, and 0.75, and then we can ditch the quantiles option.

jonathon-love:
oh, this looks great btw :)

victor-moreno:
I agree, the quartiles are redundant if it defaults to 0.25,0.5,0.75
Changing it should be careful because the cutpoints check the quartiles to avoid redundancy

jonathon-love:
cool, well ravi's jamovi day is tuesday, so if you can get it done before then, it will save you waiting a week for him :)

jonathon-love:
you definitely take an iterative approach, victor :P

victor-moreno:
Hi,
The last changes remove the quartiles and check percentiles to remove redundant ones if the same quantiles have been selected before.

jamovi/descriptives.u.yaml
```diff
@@ -77,6 +75,19 @@ children:
...
+                    name: pcVal
+                    label: Percentiles
+                    style: inline
+                    verticalAlignment: center
+                    children:
+                      - type: TextBox
+                        name: pcValues
+                        label: ''
+                        format: string
+                        enable: (pcVal)
```
raviselker: It's nicer if the width of the textbox is a bit wider. You can just add `width: large` here

R/descriptives.b.R
```diff
@@ -1016,19 +1016,39 @@ descriptivesClass <- R6::R6Class(
...
+ 
+                private$colArgs$name <- c(colArgs$name, paste0('quant', 1:(pcNEqGr-1)))
+                private$colArgs$title <- c(colArgs$title, paste0(round(pcEq * 100, 2), 'th percentile'))
+                private$colArgs$type <- c(colArgs$type, rep('number', pcNEqGr - 1))
+                private$colArgs$visible <- c(colArgs$visible, rep("(pcEqGr)", pcNEqGr - 1))
+            }
+
+            if ( self$options$pcVal ){
+
+                pcValues <- as.numeric(unlist(strsplit(self$options$pcValues,",")))
```
raviselker: I think it's nicer if the R users of `jmv` can just use a numeric vector to specify the percentile values (so `c(0.25, 0.5, 0.75)`), which means you should only do this stringsplit if `pcValues` is a string.

jamovi/descriptives.a.yaml
```diff
@@ -329,4 +321,20 @@ options:
...
+      type: Bool
+      default: false
+      description:
+          R: >
+            `TRUE` or `FALSE` (default), provide percentiles
+
+    - name: pcValues
+      title: Percentile values
+      type: String
+      default: "0.25,0.5,0.75"
```
raviselker: Why are you using proportions instead of percentages here? I would expect this to be "25, 50, 75"

R/descriptives.b.R
```diff
@@ -1065,46 +1085,79 @@ descriptivesClass <- R6::R6Class(
...
                     for (i in 1:(pcNEqGr-1))
                         stats[[paste0('quant', i)]] <- quants[i]
                 }
 
+                if ( self$options$pcVal ) {
+                    pcValues <- as.numeric(unlist(strsplit(self$options$pcValues,",")))
+                    pcValues[pcValues < 0 | pcValues > 1] <- NA 
+                    pcValues <- pcValues[!is.na(pcValues)]
+                    pcValues <- pcValues[ ! (pcValues %in% pcEq) ]
+                    npcValues <- length(pcValues)
```
raviselker: You do this a couple of times in the code, so I would maybe move this to a separate method (e.g., .convertPcValues) that stores these values into member values (see for instance line 18 in `descriptives.b.R`).

raviselker (changes requested):
It's looking good. Check out the comments for a couple of improvements that I think would be nice to add.

R/descriptives.b.R
victor-moreno: I thought pcValues will always be a string if it is so defined in u.yaml (comment before).
I will check if it is a numeric vector and avoid this split in that case

jamovi/descriptives.a.yaml
victor-moreno: This is to make it consistent with the cutoffs that will be passed to quantile(x, probs=
I will change that to percentages and divide by 100 in the code

jamovi/descriptives.a.yaml
```diff
@@ -329,4 +321,20 @@ options:
       description:
           R: >
             an integer (default: 4) specifying the number of equal groups
+
+    - name: pcVal
```
raviselker: I would rename this option to `pc`. It's easy to mix up `pcVal` and `pcValues`.

R/descriptives.b.R
```diff
@@ -1014,21 +1014,54 @@ descriptivesClass <- R6::R6Class(
...
+                pcEq <- round(pcEq, 4)
+            } else {
+                pcEq <- NULL
+            }
+
+            pcValues<-self$options$pcValues
+            if ( is.character(pcValues) )
+                pcValues <- as.numeric(unlist(strsplit(pcValues,",")))
+            pcValues <- pcValues / 100
+            pcValues[pcValues < 0 | pcValues > 100] <- NA
```
raviselker: At the moment I'm getting the following error when including a pcValue > 100:
```Error in quantile.default(column, pcValues) : 'probs' outside [0,1]```

I would try to catch the error and give an informative error message instead (e.g., "one or more "pcValues" have a value below 0 or above a 100. All "pcValues" need to be between 0 and 100.)

raviselker (changes requested):
At the moment the unit test for this analysis (tests/testthat/testdescriptives.R) doesn't work anymore because it still uses the `quart` parameter. So you need to change this as well

jamovi/descriptives.a.yaml
victor-moreno: OK I will do a global rename

R/descriptives.b.R
victor-moreno: should be pcValues >1] once it is divided by 100. I have fixed that

victor-moreno: I have changed the default pcValues in function descriptives (descriptives.h.R), but the change is overwritten and the default value for pcValues in descriptives shows as "25,50,75". Use of c(25,50,75) works in R

args(descriptives)
function (data, vars, splitBy = NULL, freq = FALSE, hist = FALSE,
    dens = FALSE, bar = FALSE, barCounts = FALSE, box = FALSE,
    violin = FALSE, dot = FALSE, dotType = "jitter", qq = FALSE,
    n = TRUE, missing = TRUE, mean = TRUE, median = TRUE, mode = FALSE,
    sum = FALSE, sd = TRUE, variance = FALSE, range = FALSE,
    min = TRUE, max = TRUE, se = FALSE, skew = FALSE, kurt = FALSE,
    sw = FALSE, pcEqGr = FALSE, pcNEqGr = 4, pc = FALSE, pcValues = **"25,50,75"**,
    formula)

I can't find where I should change this.

raviselker: Maybe @jonathon-love can chime in?

raviselker:
> At the moment the unit test for this analysis (tests/testthat/testdescriptives.R) doesn't work anymore because it still uses the `quart` parameter. So you need to change this as well

Don't forget to change the unit test :)

R/descriptives.b.R
jonathon-love: hey,

i'm not sure i understand the problem. the default value of "25, 50, 75" is correct?

with thanks

victor-moreno: This is not essential, since both jamovi and jmv R package work as expected when pcValues is defined as character "25,50,75", but probably would be better to set the default as a numeric vector c(25,50,75) in the function to be used from R.

Although I have changed the default value in function descriptives, file descriptives.h.R, I think this file is regenerated and changes have no effect on the actual code in the compiled jmv package. I don't know how the function descriptives is generated from the code to set as default for pcValues=c(25,50,75)

victor-moreno:
I have revised tests/testthat/testdescriptives.R

R/descriptives.b.R
raviselker: Yeah, it is indeed regenerated so if you commit the version of `descriptives.h.R` that you get after running `jmvtools::install()` everything is good and I'll approve the pull request. Good work!

victor-moreno: Done, I have left the code as of descriptives.h.R as generated

raviselker (approved):
It's ready to merge. I'll leave it to @jonathon-love cause we probably want to squash all those commits into one.

R/descriptives.b.R
jonathon-love: > probably would be better to set the default as a numeric vector c(25,50,75)

ah yeah, this is a limitation at this time ... (and probably not worth the effort on my part).

jonathon-love:
thanks victor! this is really great work!

PR #276, opened 2020-12-15
Bug/jmv 45 lin regression not working 10000 cases

raviselker:
Fixes jamovi/jamovi-fun-todo-list#45

### DIFF
--- a/R/linreg.b.R
+++ b/R/linreg.b.R
@@ -6,13 +6,8 @@ linRegClass <- R6::R6Class(
     active = list(
         residuals = function() {
             data <- private$.cleanData()
-            formulas <- private$.formulas()
-
-            res <- list()
-            for (i in seq_along(formulas)) {
-                model <- lm(formulas[[i]], data=data)
-                res[[i]] <- model$residuals
-            }
+            models <- private$.computeModels(data)
+            res <- private$.computeResiduals(models)
 
             return(res)
         }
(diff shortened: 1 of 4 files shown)


R/linreg.b.R
```diff
@@ -110,16 +97,66 @@ linRegClass <- R6::R6Class(
                         anovaTerms=anovaTerms, AIC=AIC, BIC=BIC, CI=CI, CIScaled=CIScaled,
                         dwTest=dwTest, VIF=VIF, cooks=cooks))
         },
+        .computeModels = function(data, model = NULL, scaled = FALSE) {
```
jonathon-love: could this be `modelNo` rather than `model`? `model` makes me expect the model object.

R/linreg.b.R
```diff
@@ -715,27 +742,26 @@ linRegClass <- R6::R6Class(
...
             }
         },
         .qqPlot = function(image, ggtheme, theme, ...) {
 
             if (is.null(image$state))
                 return(FALSE)
 
-            p <- ggplot(data=image$state, aes(x=x, y=y)) +
+            data <- private$.cleanData()
+            model <- private$.computeModels(data, model=image$state)[[1]]
```
jonathon-love: you're better off assigning the models to (something like) `private$.models` in the `.run()` phase, and only computing them here if `private$.models` is unassigned. when the analysis is run, the same analysis object is used to create all the plots. it's only when the user, say, exports an image, that the plot function is called without `.run()` being called first.

you could go further of course, have a caching active binding, etc.

jonathon-love (changes requested):

R/linreg.b.R
raviselker: So something like: 
```        
models = function() {

            if (! is.null(private$.models))
                return(private$.models)

            data <- private$.cleanData()
            models <- private$.computeModels(data)

            return(models)
},
```

R/linreg.b.R
raviselker: Fixed

R/linreg.b.R
raviselker: Fixed this by a major refactoring

jonathon-love:
just a heads up ... i think the next time we push jmv to CRAN, CRAN will insist that all these active bindings are documents :/

PR #277, opened 2020-12-16
Feature/rm anova normality assumptions

raviselker:
Fixes (partially) jamovi/jamovi#952

### DIFF
--- a/R/anovarm.b.R
+++ b/R/anovarm.b.R
@@ -52,6 +52,7 @@ anovaRMClass <- R6::R6Class(
                 private$.populateEffectsTables(result)
                 private$.populateSpericityTable(result)
                 private$.populateLeveneTable()
+                private$.prepareQQPlot(result)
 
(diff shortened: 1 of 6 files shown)


R/anovarm.b.R
```diff
@@ -702,6 +703,33 @@ anovaRMClass <- R6::R6Class(
...
+        .prepareQQPlot = function(model) {
+            image <- self$results$assump$qq
+
+            suppressMessages({
+                suppressWarnings({
+                    residuals <- residuals(model)
+                })
+            })
+
+            image$setState(residuals)
```
jonathon-love, line 715: wait a second ... i mean, just hold on a second ... what?! (i suppose you're banking on RM ANOVA data sets always being relatively small?)

raviselker: Oh yeah, I did this before you introduced me to the "don't put things in the state that grow as the dataset grows" bug. So I need to refactor this a bit :-P

PR #278, opened 2020-12-23
descriptives: IQR, CV & mean in boxplot

victor-moreno:
Hi,
This addition follows some requests: descriptives can show the IQR and the coefficient of variation. 
For CV I have used sd/mean without multiplying by 100 as is done some times.

Also, a blue dot with the mean can be added to boxplot, violin or dot plots.
Color, shape and size are fixed to make it simple.

I hope you like it
Victor

### DIFF
--- a/R/descriptives.b.R
+++ b/R/descriptives.b.R
@@ -5,15 +5,16 @@ descriptivesClass <- R6::R6Class(
     private=list(
         #### Member variables ----
         colArgs = list(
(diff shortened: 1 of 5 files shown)


victor-moreno:
I have added a second commit that adds the Mantel-Haenszel trend test for a 2xc table

raviselker:
Hi @victor-moreno, can you create a separate PR for the Mantel-Haenszel functionality? It's always best to create different PRs for different functionality.

PR #280, opened 2020-12-27
descriptives XY plot 

victor-moreno:
HI,
This new pull request adds XY plots to descriptives when at least 2 variables are added.
The plots can show dots and/or lines, regression line (linear, quadratic, cubic, log) and loess.
Also some customization of axes are possible.

Victor

### DIFF
--- a/R/descriptives.b.R
+++ b/R/descriptives.b.R
@@ -369,13 +369,33 @@ descriptivesClass <- R6::R6Class(
                         group$add(image)
                     }
 
+                    if (length(vars)>1 && var != vars[1] &&
+                        (self$options$dispersion || self$options$lineplot || self$options$linefit) ) {
+
+                        size <- private$.plotSize(levels, 'xyplot')
+
+                        image <- jmvcore::Image$new(
+                            options = self$options,
+                            name = "xyplot",
+                            renderFun = ".xyPlot",
+                            width = size[1],
+                            height = size[2],
+                            clearWith = list("splitBy", "dispersion", "lineplot", "linefit", "confband", "model",
+                                             "xmean", "ymean", "xymeandot", "xline", "yline", 
+                                             "xlog", "ylog", "xlim", "ylim","mergesplit")
+                        )
+
+                        group$add(image)
+                    }
+
                 }
 
                 plots$add(group)
             }
         },
 
-        #### Clear tables ----
+
+       #### Clear tables ----
         .clearDescriptivesTable = function(vChanges) {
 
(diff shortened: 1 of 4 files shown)


jonathon-love:
i wonder if this wouldn't be a better inclusion in the `scatr` module? @raviselker ?

raviselker:
Yeah, I need to have think about this one. It seems a bit much to add these plots to the descriptives, giving you a plot for every combination of variables. It would maybe indeed fit better in scatr (by just adding some options to the already existing scatterplot function), or maybe even the corrMatrix function.

raviselker:
Hi Victor, I had a thought about and also took a closer look and I actually kinda like this functionality in the descriptives. I would lose a couple of the options to simplify things, and at the moment you don't add every possible combination to the results (you always just use the first variable as the x-axis variable) which feels unintuitive. What do you think @jonathon-love? Another option would indeed be to extend the scatr module: https://github.com/raviselker/scatr.

jonathon-love:
i'm thinking we roll it into scatr, **but** we ship scatr with jamovi by default.

lots of teaching exercises require these plots, and i've just experienced this week students grappling with needing to install modules. some proportion of students don't notice the instructions that they need to install, and ask me why the menu option isn't visible, and when the modules are pre-installed, students follow the module installation instructions and wonder why they can't install it (because it's already installed).

so "why not then simply roll these analyses into jmv?" ... the issue is that these analyses we expect will be superseded when we develop a proper plot builder. at that point, we can stop shipping scatr with jamovi, but if people want the backwards compatibility, or for their old teaching materials to work, they can manually install scatr.

we're tremendously reluctant to add functionality that we'll later go on to supercede because we don't want to end up with the spss clutter situation, but i think this approach lets us have our sponge-cake and eat it too.

thoughts?

victor-moreno:
I agree to add these plots to scatr and ship scatr which jamovi by default. This is a good option. 
I also had the experience that students don't find their way to install modules, usually because they don't read instructions.
When I have some time I will study scatr and move the code there to get a little more customizable plots.

jonathon-love:
thanks victor, let me just get @raviselker to sign off on this.

jonathon-love:
we also might consider renaming scatr ... to `Legacy Plots` for example.

raviselker:
Yes, this all sounds good to me! In terms of a name change, this is also published on the CRAN at the moment under scatr, so not sure about that.

jonathon-love:
that's a green light victor! :)

PR #281, opened 2020-12-30
prediction intervals in linear regression

victor-moreno:
Hi, this is for Ravi.
I have added prediction intervals to the EMM table and plots.
I also have removed conditions is.null(names$plots) && is.null(names$lines)  in line 961 of linreg.b.R to allow confidence and prediction bands plotted by strata if other variable is in the term. Before ciEmm was ignored for a continuous variable if other was in the term.

Victor

### DIFF
--- a/R/linreg.b.R
+++ b/R/linreg.b.R
@@ -458,6 +458,9 @@ linRegClass <- R6::R6Class(
                         table$addColumn(name='lower', title='Lower', type='number', superTitle=paste0(self$options$ciWidthEmm, '% Confidence Interval'), visibl="(ciEmm)")
                         table$addColumn(name='upper', title='Upper', type='number', superTitle=paste0(self$options$ciWidthEmm, '% Confidence Interval'), visibl="(ciEmm)")
 
+                        table$addColumn(name='sd', title='SD', type='number', visibl="(piEmm)")
+                        table$addColumn(name='lowerpi', title='Lower', type='number', superTitle=paste0(self$options$ciWidthEmm, '% Prediction Interval'), visibl="(piEmm)")
+                        table$addColumn(name='upperpi', title='Upper', type='number', superTitle=paste0(self$options$ciWidthEmm, '% Prediction Interval'), visibl="(piEmm)")
                         nRows <- prod(nLevels)
 
                         for (k in 1:nRows) {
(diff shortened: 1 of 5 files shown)


raviselker:
Hi Victor,
What is the exact use case for the prediction intervals (this is the first time I see them mentioned)? I always aim to keep the analyses as simple as possible with still being as complete as possible. This sometimes means that options that are a bit niche are not added, not because it's difficult to add them, but because it would clutter the interface too much.

Also, I did some refactoring of the linreg function recently which means this PR now conflicts with the master branch.

victor-moreno:
Hi Ravi,
In regression you may be interested in predicting the mean, then you use the 95%CI as uncertainty. But if you are interested in predicting the spread of individual observations, then you need the predictive interval, that uses a larger standard deviation.
https://rstudio-pubs-static.s3.amazonaws.com/7024_e3a68a9b35454e74abfe15b621c50502.html
https://rpubs.com/Bio-Geek/71339

This distinction is important when teaching prediction in regression. Jamovi provides confidence intervals in several plots, but not prediction intervals.

If you agree to add this, I will revise the code to avoid conflicts.

jonathon-love:
in spss, i think you create prediction intervals with the 'save' functionality. it creates a prediction interval for each case?

do we want to take a similar approach to prediction intervals as spss? or am i misunderstanding something?

with thanks

victor-moreno:
I think that the plot is enough for now, but if predictions can be saved in a future version, then either the confidence or prediction interval should be also saved. I don't use SPSS, so I don't know what it does. In R you can get confidence and prediction intervals with predict.lm.

raviselker:
Hi Victor,
For now I think the drawbacks of adding the prediction interval (more clutter in the analyses, users getting confused about the difference between confidence interval and prediction interval) outweigh the benefits for me. However, if we get a lot of requests for this feature in the future, we can always add it at a later stage and you can do a new PR. In the future, it might be a good idea to contact me before you implement new functionality so that we can have a discussion about it before you do the work. This way, I won't feel as bad for rejecting it :-P

victor-moreno:
Agree, I keep the code in case we want it in the future. 

Do you think prediction intervals could have an option in scatr?
Just to get a plot like the last one here:
[https://rpubs.com/Bio-Geek/71339](https://rpubs.com/Bio-Geek/71339)

raviselker:
Yeah, adding the prediction intervals to scatr sounds good to me!

PR #282, opened 2021-01-02
conttables Mantel-Haenszel trend test

victor-moreno:
This contains only the Mantel-Haenszel trend test for conttables
Victor

### DIFF
--- a/R/conttables.b.R
+++ b/R/conttables.b.R
@@ -17,6 +17,7 @@ contTablesClass <- R6::R6Class(
             odds  <- self$results$odds
             gamma <- self$results$gamma
             taub  <- self$results$taub
+            mh  <- self$results$mh
 
             data <- private$.cleanData()
 
@@ -29,6 +30,7 @@ contTablesClass <- R6::R6Class(
                 nom$addColumn(index=i, name=layer, type='text', combineBelow=TRUE)
                 gamma$addColumn(index=i, name=layer, type='text', combineBelow=TRUE)
                 taub$addColumn(index=i, name=layer, type='text', combineBelow=TRUE)
+                mh$addColumn(index=i, name=layer, type='text', combineBelow=TRUE)
             }
 
             # add the row column, containing the row variable
@@ -178,6 +180,7 @@ contTablesClass <- R6::R6Class(
                 odds$addRow(rowKey=1, values=list())
                 gamma$addRow(rowKey=1, values=list())
                 taub$addRow(rowKey=1, values=list())
+                mh$addRow(rowKey=1, values=list())
 
             } else {
 
@@ -195,6 +198,7 @@ contTablesClass <- R6::R6Class(
                     odds$addRow(rowKey=i, values=values)
                     gamma$addRow(rowKey=i, values=values)
                     taub$addRow(rowKey=i, values=values)
+                    mh$addRow(rowKey=i, values=values)
                 }
             }
 
@@ -241,6 +245,7 @@ contTablesClass <- R6::R6Class(
             odds  <- self$results$odds
             gamma <- self$results$gamma
             taub  <- self$results$taub
+            mh    <- self$results$mh
 
             freqRowNo <- 1
             othRowNo <- 1
@@ -297,14 +302,24 @@ contTablesClass <- R6::R6Class(
                     else
                         exp <- test$expected
 
-                    if (self$options$taub) {
+                    if (self$options$taub || self$options$mh) {
                         df <- as.data.frame(as.table(mat))
                         v1 <- rep(as.numeric(df[[1]]), df$Freq)
                         v2 <- rep(as.numeric(df[[2]]), df$Freq)
 
-                        # this can be slow
-                        tau <- try(cor.test(v1, v2, method='kendall', conf.level=ciWidth))
+                        if (self$options$taub) {
+                            # this can be slow
+                            tau <- try(cor.test(v1, v2, method='kendall', conf.level=ciWidth))
+                        }
+                        if (self$options$mh) {
+                            if (all(dim(mat))>2)
+                                mhchi2 <- -1 # Mantel-Haenszel is only for 2xk tables
+                            else
+                                mhchi2 <- try((cor(v1, v2)^2)*(sum(df$Freq)-1))
+
+                        }
                     }
+
 
                     zP <- NULL
                     dp <- NULL
@@ -482,6 +497,19 @@ contTablesClass <- R6::R6Class(
                             t=unname(tau$statistic),
                             p=tau$p.value)
                     taub$setRow(rowNo=othRowNo, values=values)
+                }
+
+                if (self$options$mh) {
+                    if (base::inherits(mhchi2, 'try-error') || is.na(mhchi2) || mhchi2 == -1)
+                        values <- list(chi2=NaN, df='', p='')
+                    else
+                        values <- list(
+                            chi2=mhchi2,
+                            df=1,
+                            p=1-pchisq(mhchi2,1))
+                    mh$setRow(rowNo=othRowNo, values=values)
+                    if (mhchi2 == -1)
+                        mh$addFootnote(rowNo=othRowNo, 'chi2', 'One variable must be binary')
                 }
 
                 if ( ! is.null(lor)) {
--- a/R/conttables.b.R
+++ b/R/conttables.b.R
@@ -312,14 +312,12 @@ contTablesClass <- R6::R6Class(
                             tau <- try(cor.test(v1, v2, method='kendall', conf.level=ciWidth))
                         }
                         if (self$options$mh) {
-                            if (all(dim(mat))>2)
+                            if (all(dim(mat) > 2))
                                 mhchi2 <- -1 # Mantel-Haenszel is only for 2xk tables
                             else
-                                mhchi2 <- try((cor(v1, v2)^2)*(sum(df$Freq)-1))
-
+                                mhchi2 <- try((cor(v1, v2)^2) * (sum(df$Freq) - 1))
                         }
                     }
-
 
                     zP <- NULL
                     dp <- NULL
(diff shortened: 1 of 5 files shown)


R/conttables.b.R
```diff
@@ -484,6 +499,19 @@ contTablesClass <- R6::R6Class(
...
+                if (self$options$mh) {
+                    if (base::inherits(mhchi2, 'try-error') || is.na(mhchi2) || mhchi2 == -1)
+                        values <- list(chi2=NaN, df='', p='')
+                    else
+                        values <- list(
+                            chi2=mhchi2,
+                            df=1,
+                            p=1-pchisq(mhchi2,1))
+                    mh$setRow(rowNo=othRowNo, values=values)
+                    if (mhchi2 == -1)
```
raviselker: I can get the following error to appear:
```Error in if (mhchi2 == -1) mh$addFootnote(rowNo = othRowNo, "chi2", "One variable must be binary") : 
  missing value where TRUE/FALSE needed 
```

You can run the following code to reproduce (works fine with `mh = FALSE`):
```
set.seed(1337)
data <- data.frame(
    A = sample(letters[1:2], 100, replace = TRUE),
    B = sample(letters[1:3], 100, replace = TRUE),
    C = sample(letters[1:16], 100, replace = TRUE)
)

jmv::contTables(formula = ~ A:B:C, data = data, mh = TRUE)
```

R/conttables.b.R
```diff
@@ -484,6 +499,19 @@ contTablesClass <- R6::R6Class(
                     taub$setRow(rowNo=othRowNo, values=values)
                 }
 
+                if (self$options$mh) {
+                    if (base::inherits(mhchi2, 'try-error') || is.na(mhchi2) || mhchi2 == -1)
+                        values <- list(chi2=NaN, df='', p='')
+                    else
```
raviselker, line 503: If the content of one or more of the if (or else) statement contains multiple lines, always use parentheses for clearity. So here you can either make one line of the `values <- list(....)` statement or you can add parentheses to both the if and the else.

R/conttables.b.R
```diff
@@ -297,15 +302,25 @@ contTablesClass <- R6::R6Class(
...
+                        if (self$options$taub) {
+                            # this can be slow
+                            tau <- try(cor.test(v1, v2, method='kendall', conf.level=ciWidth))
+                        }
+                        if (self$options$mh) {
+                            if (all(dim(mat))>2)
+                                mhchi2 <- -1 # Mantel-Haenszel is only for 2xk tables
+                            else
+                                mhchi2 <- try((cor(v1, v2)^2)*(sum(df$Freq)-1))
+
```
raviselker: Remove empty line

jamovi/conttables.a.yaml
```diff
@@ -291,6 +291,14 @@ options:
           R: >
             `TRUE` or `FALSE` (default), provide Kendall's tau-b
 
+    - name: mh
```
raviselker, line 294: At the moment the option for Mantel-Haenszel is displayed underneath the "ordinal" heading. Looking at the wikipedia it seems you can also use it for nominal data. If this is the case, please move it to the "nominal" heading.

jamovi/conttables.r.yaml
```diff
@@ -322,3 +322,27 @@ items:
           title: p
           type: number
           format: zto,pvalue
+
+    - name: mh
+      title: Mantel-Haenszel test for trend
```
raviselker: Title of tables in header case. So: "Mantel-Haenszel Test for Trend"

R/conttables.b.R
```diff
@@ -297,15 +302,25 @@ contTablesClass <- R6::R6Class(
...
+                        }
                     }
 
+
```
raviselker: Remove empty line

R/conttables.b.R
```diff
@@ -297,15 +302,25 @@ contTablesClass <- R6::R6Class(
...
                         v2 <- rep(as.numeric(df[[2]]), df$Freq)
 
-                        # this can be slow
-                        tau <- try(cor.test(v1, v2, method='kendall', conf.level=ciWidth))
+                        if (self$options$taub) {
+                            # this can be slow
+                            tau <- try(cor.test(v1, v2, method='kendall', conf.level=ciWidth))
+                        }
+                        if (self$options$mh) {
+                            if (all(dim(mat))>2)
```
raviselker: Add spaces around logical operators. So: `...dim(mat)) > 2....`

R/conttables.b.R
```diff
@@ -297,15 +302,25 @@ contTablesClass <- R6::R6Class(
...
+                                mhchi2 <- -1 # Mantel-Haenszel is only for 2xk tables
+                            else
+                                mhchi2 <- try((cor(v1, v2)^2)*(sum(df$Freq)-1))
```
raviselker: In most cases, surround mathematical operators by spaces. See https://style.tidyverse.org/syntax.html#infix-operators for more information. This whole style guide would be good to go through at some point. Adhering to a good style guide will make code easier to read.

raviselker (changes requested):
Hi Victor, Excellent addition to the conttables! I added a couple of comments, but after you address these, I'll accept this PR.

jamovi/conttables.a.yaml
victor-moreno: Hi, 
This is the Mantel-Haenszel test for trend and only makes sense for ordinal data. I couldn't find this in wikipedia. Wikipedia refers to a different Cochran-Mantel-Haenszel statistic to calculate an OR pooled over strata of a third categorical variable.

raviselker: Ah, I just assumed this was the same statistic. You can ignore this comment then :)

R/conttables.b.R
victor-moreno: Hi Ravi
I have fixed that, and also have cached two more bugs, one detecting tables >2x2 and other with the footnote, that was not shown because was declared before the output table. Also have formated df as integer.
Together with the coding style fixes, that I appreciate, I think it should be ready now.

victor-moreno:
Hi, 
I discovered today that vcd package used to calculate log-odds-ratios and from them the OR provide wrong results when there is a zero in the 2x2 table. It should be zero or Inf, but the result is a variable number near to zero or large, but not the correct answer.

```
df <- data.frame(
A = factor(c(0,0,1,1)),
B = factor(c(0,1,0,1)),
count = c(34,76,12,0) )

data <- df[rep(seq_len(nrow(df)), df$count), 1:2]
jmv::contTables(data=data, formula= ~ A:B, logOdds = TRUE, odds = TRUE)

 ──────────────────────────────
   A        0     1     Total
 ──────────────────────────────
   0        34    76      110
   1        12     0       12
   Total    46    76      122
 ──────────────────────────────

                    Value         Lower          Upper
 ────────────────────────────────────────────────────────────
   Log odds ratio     -4.015207      -6.870342    -1.160072
   Odds ratio        0.01803922    0.001038122    0.3134635
────────────────────────────────────────────────────────────
```


I propose to change the code to use fisher.test function, that doesn't require any additional library and provides an exact confidence interval. From that we can calculate the log-OR.

If you agree, I can prepare the changes and profit this PR

raviselker:
So it looks like vcd applies a continuity correction when there are zero cells:

> **correct:** logical or numeric. Should a continuity correction be applied before computing odds? If TRUE, 0.5 is added to all cells; if numeric (or an array conforming to the data) that value is added to all cells. By default, this not employed unless there are any zero cells in the table, but this correction is often recommended to reduce bias when some frequencies are small (Fleiss, 1981).

Wouldn't this be an adequate solution to the problem of zero count cells?

victor-moreno:
I have been reading about that and the best answer is not an easy decision. If you need a standard error of the OR or logOR for meta-analysis, then adding 0.5 is required. However, to get only the point estimate and 95% CI the exact method can be used. This paper recommends it whenever possible:
https://besjournals.onlinelibrary.wiley.com/doi/full/10.1111/j.2041-210x.2012.00250.x

I suppose we can leave the code as it is. Perhaps a footnote could be added when there is a zero in the table to warn that the correction was applied.

raviselker (approved):
It all looks good. And I like the idea for the footnote in case of zero counts. Can you have a look at that?

victor-moreno:
I have added the footnote. The text is 'Haldane-Ascombe correction applied'. 

I found the reference to the original suggestion of adding 0.5 in the prior to the last paragraph sentence here  https://besjournals.onlinelibrary.wiley.com/doi/full/10.1111/j.2041-210x.2012.00250.x
I think adding the reference is not needed. If you paste Haldane-Ascombe correction in google you get this reference on second position.

Also, I have adde a condition to avoid computing estimators in 2x2 table if table in fact is 2x1 or 1x2 because a row or column is all zeros. Since 0.5 is added to each cell, now you get an artificial result for OR and logOR though the chi-square test is NaN.

R/conttables.b.R
```diff
@@ -484,6 +497,20 @@ contTablesClass <- R6::R6Class(
...
+                if (self$options$mh) {
+                    if (base::inherits(mhchi2, 'try-error') || is.na(mhchi2) || mhchi2 == -1)
+                        values <- list(chi2=NaN, df='', p='')
+                    else
+                        values <- list(chi2=mhchi2, df=1, p=1-pchisq(mhchi2,1))
+                    
+                    mh$setRow(rowNo=othRowNo, values=values)
+                    
+                    if (base::inherits(mhchi2, 'try-error') || is.na(mhchi2))
+                        mh$addFootnote(rowNo=othRowNo, 'chi2', 'Variables must have >=2 categories')
```
jonathon-love: `Variables must have at least two levels`?

R/conttables.b.R
```diff
@@ -484,6 +497,20 @@ contTablesClass <- R6::R6Class(
...
+                    else if (mhchi2 == -1)
+                        mh$addFootnote(rowNo=othRowNo, 'chi2', 'One variable must be binary')
```
jonathon-love: `At least one variable must have two levels`?

jonathon-love (commented):
thanks victor, a couple of suggestions from me now.

victor-moreno:
Done

PR #284, opened 2021-01-02
descriptives -IQR and CV added

victor-moreno:
This computes IQR and CV

### DIFF
--- a/R/descriptives.b.R
+++ b/R/descriptives.b.R
@@ -5,15 +5,16 @@ descriptivesClass <- R6::R6Class(
     private=list(
         #### Member variables ----
         colArgs = list(
-            name = c("n", "missing", "mean", "se", "median", "mode", "sum", "sd", "variance", "range",
+            name = c("n", "missing", "mean", "se", "median", "mode", "sum", "sd", "variance", "iqr", "cv","range",
                      "min", "max", "skew", "seSkew", "kurt", "seKurt", "sww", "sw"),
             title = c("N", "Missing", "Mean", "Std. error mean", "Median", "Mode", "Sum", "Standard deviation", "Variance",
+                      "IQR", "Coeff. variation", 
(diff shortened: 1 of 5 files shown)


raviselker:
Hi Victor,
Thanks for this PR. I can see the use for IQR in descriptives, but I never heard of CV before. Was there a feature request for CV somewhere? It's pretty easy to calculate this using the mean and standard deviation, so if it's a bit of niche functionality I would rather leave it out of the descriptives (because of cluttering of options).

victor-moreno:
Hi Ravi,
The coefficient of variation is useful to compare dispersion of variables with different means. It is true that it is easy to calculate, but if you have several variables and want to check which one is more or less dispersed, the CV will give the answer. 
Also, it is often used as indicator of the precision of a measure or repeatability of an assay.

jonathon-love:
hey victor,

i think let's merge IQR, and we put CV into a separate module ... perhaps an addon for descriptives? i'm happy to help you out with that. (i've never heard of CV either).

jonathon

victor-moreno:
OK to leave out the CV, perhaps the addon module can include the XY plots also. I studied the moretests module and understand that a module can modify the output, but I don't know if a module over jmv can also modify the ui to let the user select additional analyses. That would help a lot
Victor

jonathon-love:
> I don't know if a module over jmv can also modify the ui

yeah, that's a current short-coming. the plan is for addon modules to get their own expander inside the existing analyses ... but we're not there yet. :/

i'll just cc @dropmann because that's his area, but i don't think it's on the short-term roadmap. all these cool things we'd like to do. they take time unfortunately.

with moretests, i just piggy-backed off existing options (i.e. add kolmogorov-smirnoff, when the user requests shapiro-wilk) ... but i don't think there's any analagous options to use in this case.

jonathon

jonathon-love:
merged IQR

PR #297, opened 2021-03-23
Feature/jam 44 improve error reporting

raviselker:


R/errors.R
```diff
@@ -0,0 +1,82 @@
...
+variableContainsMissing = function(col, colName) {
+    fail <- any(is.na(col))
+    if (fail) {
+        jmvcore::reject(
+            jmvcore::format("'{}' contains missing values", colName),
+            code = checkTypes$CONTAINS_MISSING,
+            expected = TRUE
+        )
+    }
+}
```
jonathon-love: we don't distinguish between missing and NaN ... and NaN isn't a term which our users will necessarily be familiar with

R/errors.R
```diff
@@ -0,0 +1,82 @@
+checkTypes <- list(
+    "CONTAINS_INF" = "variable_contains_inf",
+    "CONTAINS_NAN" = "variable_contains_nan",
+    "CONTAINS_MISSING" = "variable_contains_missing",
+    "CONTAINS_ONLY_MISSING" = "variable_contains_only_missing",
+    "CONTAINS_ONE_UNIQUE_VALUE" = "variable_contains_one_unique_value"
+)
```
jonathon-love, line 6: is there much value in assigning these to a list, rather than just using `"variable_contains_inf"` directly?

R/descriptives.b.R
```diff
@@ -774,6 +774,8 @@ descriptivesClass <- R6::R6Class(
                 return(FALSE)
 
             data <- image$state$data
+            checkData(data, checkTypes$CONTAINS_ONLY_MISSING)
```
jonathon-love: i actually like the current behaviour in descriptives where it reports an N of zero, and `NaN` for all the results

R/errors.R
raviselker: Fixed

R/errors.R
raviselker: Yeah, you would want a list (or some kind of enum) that lists all the possible errors that are supported at the moment, so I would leave that in.

R/descriptives.b.R
raviselker: Yeah, the histogram wasn't working on empty data. But yeah, this is not the best solution so I took this out of the commit.

R/errors.R
jonathon-love: so a key/value system just adds another layer of redirection ... why not simply use an array of values?

raviselker: I need to be able to refer to them someway when I'm passing them them to the `checkData` function. This way, I can just refer to the key (instead of some index) from the `checkTypes` and I know that'll work. Would of course be great if R supported some kind of type hints so this would be even more valuable, but alas

jonathon-love: make the key/value the same?

raviselker: How about the key is the same but in uppercase? I'm used to making constants and enum keys uppercase...

jonathon-love: my preference in all of this is to use R conventions, so i would favour lowercase, ... but if it's not going to visible outside of jmv then i suppose it doesn't matter

raviselker: yeah, I *was* wondering whether this should go into jmvcore...

jonathon-love: in which case i'd push you towards lowercase

raviselker: yeah, sure. I'll change it. And let's put it into `jmv` for now and then move it at a later point, no?

jonathon-love: yup, sure

raviselker: k, fixed

PR #299, opened 2021-03-28
Test value for independent and paired samples

arcaldwell49:
The "test value" option for the other t-test types (independent and paired). The footnote for the t-test tables has also been changed to better reflect the actual hypothesis tests.

Please note: the Bayes Factor for the independent samples t-test does not allow for a non-zero "mu" argument.  For now it will print a warning as a footnote for this table. However, I do know that this is possible based on the work of Don van Ravenzwaaij https://bmcmedresmethodol.biomedcentral.com/articles/10.1186/s12874-019-0699-7

I will look into modifications to the BayesFactor package tomorrow and get in contact with Richard Morey if necessary to get an update to that package.

Please feel free to reach out for any comments or questions. I am also happy to work on this further if you encounter any issues.

Lovely work on jamovi all. Cheers!

### DIFF
--- a/R/ttestis.b.R
+++ b/R/ttestis.b.R
@@ -18,6 +18,8 @@ ttestISClass <- R6::R6Class(
                 data[[name]] <- jmvcore::toNumeric(data[[name]])
             data[[groupVarName]] <- droplevels(as.factor(data[[groupVarName]]))
 
+            testValue <- self$options$get("testValue")
+
             ttestTable <- self$results$ttest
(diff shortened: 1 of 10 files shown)


jonathon-love:
hi aaron,

my feeling is we won't end up merging this, sorry. i'd need to discuss it with the team.

jonathon

jonathon-love:
i'd welcome changes to the hypothesis footnotes however.

PR #305, opened 2021-05-20
linReg: fix "differing number of rows" issue with fitted values

raviselker:


R/linreg.b.R
```diff
@@ -889,9 +903,9 @@ linRegClass <- R6::R6Class(
             }
 
             if (self$options$predictOV && self$results$predictOV$isNotFilled()) {
-                self$results$predictOV$setRowNums(names(self$fitted[[1]]))
+                self$results$predictOV$setRowNums(names(self$predicted[[1]]))
                 for (i in seq_along(self$fitted))
```
jonathon-love, line 907: you don't want to do all this in a single function call? you can just call predictOV$set(...) once rather than requiring a loop, etc.

raviselker: Yeah, I could do that but it seems a different issue from what this PR is solving

jonathon-love: yeah, i don't doubt that this PR solves the problem, but it's standard to request that a PR reflect good practice as well. ... but i don't really care and am happy to merge.

raviselker: Oh yeah, it was more that this code was not introduced in this PR so maybe something that should be fixed outside this PR

araappa: Hi, and sorry, I'm not sure if this issue is still open. However, I'll get this "arguments imply differing number of rows:"-error when using jamovi 2.2.5.0 and Correlation Coefficients in MAJOR analysis. Could you kindly give me your advice on this, thank you!

jonathon-love: i'd suggest posting this in the forums, and i'll see if i can attend maurizio to it.

https://forum.jamovi.org

PR #322, opened 2021-10-26
contTables: fixed spaces in var names bug

raviselker:
* Error would appear when barplots are requested and variable names have spaces
* Added test for this bug
* Fixed test for anovaRMNP

Might conflict with #321


jonathon-love:
i split this commit in two, so i could backport the fix to *before* the i18n stuff landed (so we can push out a 2.2.3).

so this is the first commit:

https://github.com/jamovi/jmv/commit/07a04d8c8bc211042b1966590b54af73197a4ee9

and then i folded this mod into the i18n commit:

https://github.com/jamovi/jmv/commit/67bd7ef047bb66c594d45d9bc72e3a6a2d5eb07d#diff-bdf88e4bc4e6c16d4abfa216332dac46482f9b14cc7326d01f022aa8126f7abbR640

PR #328, opened 2021-12-13
Added Spanish

victor-moreno:
Spanish translation of jmv
The R example runs/outputs have not been translated, just copied from the English version

Also I found that the y -axis of the Histogram and Density of Descriptives always uses the label "density". I will check if that requires revising the ggplot code.

This is a very long file and may still have some problems. I have checked the UI and generally looks good, but I have not tested all outputs.

### DIFF
--- a/jamovi/i18n/es.po
+++ b/jamovi/i18n/es.po
@@ -1,0 +1,7280 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: \n"
+"POT-Creation-Date: 2021-07-26 12:08:06+01000\n"
+"PO-Revision-Date: 2021-12-12 14:49+0100\n"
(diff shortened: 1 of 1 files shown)


jonathon-love:
> This is a very long file and may still have some problems. I have checked the UI and generally looks good, but I have not tested all outputs.

no worries ... some kinks are to be expected.

jonathon-love:
hi victor,

we've added what are hopefully the last strings before the 2.3 release. could you take a look when you've got a moment.

with thanks

jonathon-love:
https://github.com/jamovi/jamovi-i18n/blob/main/src/es.po

https://github.com/jamovi/jmv/blob/master/jamovi/i18n/es.po

PR #330, opened 2022-01-04
efa: added factor scores

raviselker:


raviselker:
@jonathon-love there's still two issues:

1. The `factorScoreMethod` cannot be enabled based on the value of `factorScoresOV`
2. The description of the factor score variables doesn't update when you select a different `factorScoreMethod`

jonathon-love:
> The description of the factor score variables doesn't update when you select a different factorScoreMethod

i think you're missing a `clearWith`

jonathon-love:
oh no, you've used `clearWith` appropriately.

so we don't overwrite the description of a column in case the user has changed it ... to date, a change in description has always been accompanied by a change in name, and a change in name (by the analysis) triggers an overwriting of the name and the description. here is the updating implementation:

https://github.com/jamovi/jamovi/blob/current-dev/server/jamovi/server/instance.py#L382-L384

we could add an `output_desired_column_description` to the column class, and only update the column description from the output object if it matched it.

PR #334, opened 2022-01-08
Added Chinese Simplified translation

mengtegroup:


shun2wang (commented):
Great! Thank you for your work！

There some suggestions for modification:

- Should distinguish between Bayesian Credible interval（可信区间） and confidence interval（置信区间）,In many cases, the confidence interval calculated from Bayesian or frequency is the same in value, but the meaning is different，see [here](https://blog.csdn.net/TSzero/article/details/111676121) 
- N can usually not be translated or written as ‘计数’， '例数' used for medical statistics but not universal,
- English spaces should be used for spacing in mixed Chinese and English/code/punctuation sentences，such as `因子 '{factorName}' 不包含任何数据` but not `因子'{factorName}'不包含任何数据`
- I think AIC/BIC....etc. abbreviation String should Put after Chinese in a'()'，Or do not translate it.

mengtegroup:
Thank you for your suggestions! We've modified accordingly and opened a new PR.

PR #338, opened 2022-02-01
Update ja.po

sbtseiji:
Added translations for new words


sjentsch:
We can perhaps close this issue? It is implemented.

PR #339, opened 2022-02-10
descriptives: add outlier labels option for box plots

raviselker:
* Added the following dependencies: 'ggrepel', 'tibble', 'dplyr', and 'magrittr'


R/descriptives.b.R
```diff
@@ -1128,6 +1129,13 @@ descriptivesClass <- R6::R6Class(
                 else
                     x <- names$s1
 
+                if (self$options$box && self$options$boxLabelOutliers) {
+                    data <- data %>%
+                        tibble::rownames_to_column('.ROWNAMES') %>%
```
jonathon-love: i appreciate `tibble` exists in the bowels of our dependencies anyway ... but i totally resent the hadleyverse ... can we drop this explicit dependency?

raviselker: As in, don't use tibble, dplyr, and magrittr? Because these things make plotting a lot easier tbh

jonathon-love: yeah, i'm only proposing to drop `tibble`

raviselker: Just drop the `tibble` dependency, or the complete use?

jonathon-love: so if we changed this:

```
data <- data %>%
    tibble::rownames_to_column('.ROWNAMES') %>%
```

to

```
data$.ROWNAMES <- rownames(data)
data <- data %>%
```

then we can ditch the `tibble` import

raviselker: ok, I removed the dependency

PR #341, opened 2022-03-09
Fix to term issues

jonathon-love:
Fixes jamovi/jamovi#1181


raviselker (approved):
This all seems fine (no idea why I only used single brackets before)

jonathon-love:
> no idea why I only used single brackets before

it's hard to find good help

PR #342, opened 2022-03-16
Update descriptives.b.R

atpinto:
Changed the critical value for the confidence interval for the mean. The critical values is now chosen as the quantile from the t-distribution with df=n-1 rather than the quantile from the normal distribution.

### DIFF
--- a/R/descriptives.b.R
+++ b/R/descriptives.b.R
@@ -1417,8 +1417,10 @@ descriptivesClass <- R6::R6Class(
                 stats[['min']] <- min(column)
                 stats[['max']] <- max(column)
                 stats[['se']] <- sqrt(var(column)/length(column))
-                zQuant <- 1 - ((1 - self$options$ciWidth/100) / 2)
-                ciDiff <- qnorm(zQuant) * stats[['se']]
+                #zQuant <- 1 - ((1 - self$options$ciWidth/100) / 2)
+                tQuant <- 1 - ((1 - self$options$ciWidth/100) / 2)
+                #ciDiff <- qnorm(zQuant) * stats[['se']]
+                ciDiff <- qnorm(tQuant, df=length(column)-1 ) * stats[['se']]
                 stats[['ciLower']] <- stats[['mean']] - ciDiff
                 stats[['ciUpper']] <- stats[['mean']] + ciDiff
                 stats[['iqr']] <- diff(as.numeric(quantile(column, c(.25,.75))))


atpinto:
The confidence interval for the mean in the Descriptive statistics command is using the critical values from the normal distribution instead to the t-distribution.  When the sample size is small size, this confidence interval will fail to have the correct coverage because is too narrow.   Given that we are using the sample standard deviation for the calculation of the standard error, the t-distribution should be used for the calculation of the confidence interval for the mean.

The correct confidence interval can be obtained form the t-test command and it will not match the normal-based  confidence interval in the descriptives command.

R/descriptives.b.R
```diff
@@ -1417,8 +1417,10 @@ descriptivesClass <- R6::R6Class(
                 stats[['min']] <- min(column)
                 stats[['max']] <- max(column)
                 stats[['se']] <- sqrt(var(column)/length(column))
-                zQuant <- 1 - ((1 - self$options$ciWidth/100) / 2)
-                ciDiff <- qnorm(zQuant) * stats[['se']]
+                #zQuant <- 1 - ((1 - self$options$ciWidth/100) / 2)
```
raviselker, line 1420: Please remove commented out line
```suggestion
```

R/descriptives.b.R
```diff
@@ -1417,8 +1417,10 @@ descriptivesClass <- R6::R6Class(
...
+                tQuant <- 1 - ((1 - self$options$ciWidth/100) / 2)
+                #ciDiff <- qnorm(zQuant) * stats[['se']]
```
raviselker, line 1422: Please remove commented out line
```suggestion
```

R/descriptives.b.R
```diff
@@ -1417,8 +1417,10 @@ descriptivesClass <- R6::R6Class(
...
+                ciDiff <- qnorm(tQuant, df=length(column)-1 ) * stats[['se']]
```
raviselker, line 1423: Improve formatting
```suggestion
                ciDiff <- qnorm(tQuant, df=length(column) - 1) * stats[['se']]
```

raviselker (changes requested):
Sorry for the delay in the review, we were discussing internally for a bit wether we wanted accept this change and we came to the conclusion that it would indeed be better to use a t-distribution. So after the requested changes are applied, I'll approve this PR.

jonathon-love:
i think we want to add a footnote too @raviselker ? 1. so people are aware of the change. 2. because it's informative to have?

raviselker:
I fixed up this PR in #348 (there were some other issues with it).

PR #351, opened 2022-07-09
MANCOVA: catch mshapiro.test() error silently

MAgojam:
Captured the Shapiro-Wilk test error for cases> 5000 and put a check mark in the code so it doesn't interrupt the rest of the calculations.
A footnote in the "Shapiro-Wilk Multivariate Normality Test" output table to explain the NaN for the statistics.


jonathon-love:
thanks so much mz!

PR #357, opened 2022-08-21
Descriptives: add extreme values table

### LINKED ISSUE #168: Adding extreme values table to descriptives
Whenever a student comes in asking how to interpret a 4- or 5-way interaction in ANOVA, the first thing I do is run SPSS' Explore procedure and ask for the Extreme Values table on their dependent variable. It simply shows the 5 largest and smallest values (and case number), and very likely one or two of those values is a huge data entry error. Fixing that usually removes the high-level interaction. Such a table would be a very helpful addition to jmv's descriptives.

image

raviselker:
Table shows the extremest values (highest and lowest) of a numeric variable. The number of required extreme values can be defined by the user.

Important: this implementation ignores the grouping variable

Closes #168 

The implementation looks as follows:
Screenshot 2022-08-21 at 13 22 28

### DIFF
--- a/R/descriptives.b.R
+++ b/R/descriptives.b.R
@@ -702,8 +702,6 @@ descriptivesClass <- R6::R6Class(
             tables <- self$results$extremeValues
             vars <- self$options$vars
 
-            note <- .('Number of requested extreme values is higher than the number of rows in the data.')
-
             for (i in seq_along(vars)) {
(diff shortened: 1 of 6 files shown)


jonathon-love:
hmm, i'm wondering if there's a more natural way of saying `Extremest` ... that's also compact. ... `Most extreme`? `Min/max`?

PR #359, opened 2022-09-06
Minor: fix typo

shun2wang:
fix typo when i find in translation work, minor....

also strings in https://github.com/jamovi/jmv-i18n should be sync updated.


jonathon-love:
thanks!

PR #365, opened 2022-11-24
Update 00refs.yaml

sjentsch:
Corrected typos. The missing commas cause the latexify-script to break.


jonathon-love:
> The missing commas cause the latexify-script to break.

oh yup, but we need to be robust to missing commas? i.e. if someone uses a different module, which provides its own reference which is broken?

PR #366, opened 2022-11-27
Add github action that runs a R CMD check

raviselker:
Also fixes a couple of issues that made the R CMD check fail.


jonathon-love:
i've avoided PMCMRplus because of the number of dependencies ... could you take a look at that? happy to move across if you think the number of dependencies is reasonable.

raviselker:
I reverted the PMCMRplus commit, and it still passed the R CMD check indeed. The only thing is that if someone runs the non-parametric rm anova in `jmv`, and the PMCMR package is up to date they get a deprecation warning. This is not very nice. But if we want to fix that, I don't think that fix has to go into this MR...

jonathon-love:
merged

PR #368, opened 2022-11-30
tTests: always show hypothesis footnote

raviselker:
Previously, the hypothesis footnote was not shown when the hypothesis was the default one. However, I think it's good to always show this in the main table so that the output is more self contained.


jonathon-love:
not sure i agree, but don't care that much, so will merge this.

PR #369, opened 2023-01-19
Updated corrmatrix to correlate text-type ordinal variables as well

MAgojam:
Some changes to corrmatrix that allow correlation to be obtained for text-type ordinal variables as well.

### DIFF
--- a/R/logregbin.b.R
+++ b/R/logregbin.b.R
@@ -225,7 +225,7 @@ logRegBinClass <- if (requireNamespace('jmvcore')) R6::R6Class(
         .computeFitted = function() {
             fitted <- list()
             for (i in seq_along(self$models))
-                fitted[[i]] <- self$models[[i]]$fitted.values
+                fitted[[i]] <- ifelse(self$models[[i]]$fitted.values > 0.5, 1, 0)
 
             return(fitted)
         },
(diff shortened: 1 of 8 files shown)


jonathon-love:
looks good mz!

i assume the stats for pearson come through as `NaN` when a text variable is provided, can we provide a footnote as well, explaining why? `Pearson correlation cannot be calculated for ordered text values`? (wrap it in a `.( ... )` so it can be translated).

also your editor seems to be inserting whitespace ... can you change this so that empty lines contain no writespace?

with thanks

R/logregbin.b.R
```diff
@@ -234,7 +234,7 @@ logRegBinClass <- if (requireNamespace('jmvcore')) R6::R6Class(
             predicted <- list()
             for (i in seq_along(self$models))
                 predicted[[i]] <- predict(self$models[[i]], data, type="response")
-
+
```
jonathon-love, line 237: your text editor is still sneaking spaces in!

MAgojam: > also your editor seems to be inserting whitespace ... 

Fixed (bug in script)

> i assume the stats for pearson come through as NaN when a text variable is provided

Originally, if at least one variable was Ordinal (both numeric and text types), I had passed a NaN to Pearson with this footnote:
`matrix$setNote('mto', .("Pearson's correlation cannot be computed for ordinal-type measures"))`
I pushed a PR without, because I wanted to ask you if you agreed or preferred only for the ordinal text.
I think NaN for the pearson should be there if the variable is an ordinal type measure, whether numeric or text.

Let me know and thanks for your time,
mz

jonathon-love: NaN in pearson with the accompanying footnote would be splendid!

although i realise now that you've committed the logregbin stuff to the same branch as the corr matrix stuff. can i get you to commit these to separate branches, and open separate PRs?

PR #371, opened 2023-01-24
Update for the correlation of text-type ordinal measures

MAgojam:
Nonparametric correlation is possible if the ordinal variable is also a text-type measure.
If at least one variable is ordinal with the Pearson correlation option, returns NaN and a footnote explains this.

### DIFF
--- a/R/corrmatrix.b.R
+++ b/R/corrmatrix.b.R
@@ -14,7 +14,7 @@ corrMatrixClass <- R6::R6Class(
         for (i in seq_along(vars)) {
             var <- vars[[i]]
 
-            if(is.ordered(self$data[[var]]) && self$options$pearson)
+            if (is.factor(self$data[[var]]) && self$options$pearson)
                 mtord <- TRUE
 
             matrix$addColumn(name=paste0(var, '[r]'), title=var,
--- a/R/corrmatrix.b.R
+++ b/R/corrmatrix.b.R
@@ -10,8 +10,12 @@ corrMatrixClass <- R6::R6Class(
         nVars <- length(vars)
         ciw <- self$options$ciWidth
 
+        mtord <- FALSE
         for (i in seq_along(vars)) {
             var <- vars[[i]]
+
+            if(is.factor(self$data[[var]]) && self$options$pearson)
+                mtord <- TRUE
 
             matrix$addColumn(name=paste0(var, '[r]'), title=var,
                 type='number', format='zto', visible='(pearson)')
@@ -92,6 +96,11 @@ corrMatrixClass <- R6::R6Class(
         if ( ! flag)
             matrix$setNote('flag', NULL)
 
+        if (mtord)
+            matrix$setNote('mtord', .('Pearson correlation cannot be calculated for ordered values'))
+        else
+            matrix$setNote('mtord', NULL)
+
     },
     .run=function() {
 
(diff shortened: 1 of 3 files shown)


R/corrmatrix.b.R
```diff
@@ -92,6 +96,11 @@ corrMatrixClass <- R6::R6Class(
         if ( ! flag)
             matrix$setNote('flag', NULL)
 
+        if (mtord)
+            matrix$setNote('mtord', .('Pearson correlation cannot be calculated for ordered values'))
```
shun2wang, line 100: Could you mark this as translatable strings ?

jonathon-love: the `.( ... )` designates it as a translatable string sw

shun2wang: ah, I see. thanks

R/corrmatrix.b.R
```diff
@@ -10,9 +10,13 @@ corrMatrixClass <- R6::R6Class(
         nVars <- length(vars)
         ciw <- self$options$ciWidth
 
+        mtord <- FALSE
         for (i in seq_along(vars)) {
             var <- vars[[i]]
 
+            if(is.factor(self$data[[var]]) && self$options$pearson)
```
jonathon-love: so take a look here:

https://dev.jamovi.org/tuts0202-handling-data.html

so our approach is to treat numeric values (regardless of whether they're factors or ordinal) as continuous ... so i think this needs to be `is.character()` rather than `is.factor()` ... so NaN values are should only be generated for variables of type text.

makes sense?

R/corrmatrix.b.R
```diff
@@ -119,7 +128,12 @@ corrMatrixClass <- R6::R6Class(
                     colVarName <- vars[[j]]
                     colVar <- jmvcore::toNumeric(self$data[[colVarName]])
 
-                    result <- private$.test(rowVar, colVar, hyp)
+                    if(is.factor(self$data[[rowVarName]]) || is.factor(self$data[[colVarName]]))
+                        mtord <- TRUE
+                    else
+                        mtord <- TRUE
```
jonathon-love: i notice that `mtord` is assigned `TRUE` in both the "if true" and "if false" conditions?

R/corrmatrix.b.R
```diff
@@ -10,9 +10,13 @@ corrMatrixClass <- R6::R6Class(
         nVars <- length(vars)
         ciw <- self$options$ciWidth
 
+        mtord <- FALSE
         for (i in seq_along(vars)) {
             var <- vars[[i]]
 
+            if(is.ordered(self$data[[var]]) && self$options$pearson)
```
jonathon-love: so take a look here:

https://dev.jamovi.org/tuts0202-handling-data.html

so our approach is to treat numeric values (regardless of whether they're factors or ordinal) as continuous ... so i think this needs to be is.character() rather than is.ordered() ... so NaN values are should only be generated for variables of type text.

makes sense?

MAgojam: >makes sense?

hey,
barring my possible misunderstanding, we had discussed the possibility of allowing the correlation by ordinal text variables for the two non-parametric statistics and having a NaN for pearson if the option was activated.
Obviously a footnote to the table was necessary to prove the presence of NaN.
In a first instance it was only related for an ordinal text variable, but I suggested that the NaN to pearson could be corrected when at least one of the two was ordinal (regardless of whether it was text or integer).
It seemed to me that you either agreed with my suggestion or there was my misunderstanding here).
With is.ordered() I intercept the variable and if pearson is also true it will output NaN and the footnote.

jonathon-love: sorry, i've just realised i've not been thinking about this correctly. i thought there was a difference in R between an ordinal variable of type text, and an ordinal variable of type integer ... but that's evidently not the case.

i just need to think about this a bit more. i'll come back to you by tomorrow.

thanks for your efforts and patience!

jonathon-love: OK, i think the correct test here is `is.factor()` rather than `is.ordered()`

i think the reason the tests were failing before was because of the

```
if (...)
    mtord <- TRUE
else
    mtord <- TRUE
```

block, but now that's fixed i think this will pass the tests with `is.factor()`.

PR #373, opened 2023-01-24
CFA: catch unclear error message 

raviselker:
Error message indicates that model did not converge, but gives a non-comprehensible message instead.
This commit, introduces a way to catch the error message, and throw a more user-friendly error
message instead.

Please give this one a thorough review @jonathon-love because I do introduce some new way of error handling here.

Fixes jamovi/jamovi#1204
Fixes jamovi/jamovi#1292

### DIFF
--- a/R/cfa.b.R
+++ b/R/cfa.b.R
@@ -3,13 +3,63 @@
 cfaClass <- R6::R6Class(
     "cfaClass",
     inherit = cfaBase,
+    #### Active bindings ----
+    active = list(
+        dataProcessed = function() {
+            if (is.null(private$.dataProcessed))
+                private$.dataProcessed <- private$.cleanData()
+
+            return(private$.dataProcessed)
+        },
+        model = function() {
+            if (is.null(private$.model))
+                private$.model <- private$.computeModel()
+
+            return(private$.model)
+        },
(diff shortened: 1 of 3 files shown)


jonathon-love:
should i be concerned about 'all checks have failed'?

raviselker:
> should i be concerned about 'all checks have failed'?

Concerned no, but you should definitely not merge this PR (and other PRs for that matter) before they succeed so I'll take a look at what's happening here.

shun2wang:
maybe tests failures related to the new version of lavaan(0.6-13)? Because lavaan has recently updated and some changes may related to the calculation results .

see: bug fixes of the lavaan [change logs](https://lavaan.ugent.be/history/dot6.html)

raviselker:
> maybe tests failures related to the new version of lavaan(0.6-13)? Because lavaan has recently updated and some changes may related to the calculation results .
> 
> see: bug fixes of the lavaan [change logs](https://lavaan.ugent.be/history/dot6.html)

Only the test added in this PR failed. There were some non-ASCII characters in an error message comparison that didn't work in the test runner (but they did work locally). So I changed the error message into a partial match instead of a full match. Checks now all succeed

PR #375, opened 2023-02-03
Update: Tjur's R2 in Model Fit

MAgojam:
Jth,
It might be interesting to have a look here for the suggestion to add Tjur's one to the 3 pseudo R2s
already present.
[https://statisticalhorizons.com/r2logistic/](url)
I personally agree with the last comment by E. Shtatland (2022/12/19).

### DIFF
--- a/R/logregbin.b.R
+++ b/R/logregbin.b.R
@@ -288,8 +288,10 @@ logRegBinClass <- if (requireNamespace('jmvcore')) R6::R6Class(
                 r2mf <- 1 - dev/nullDev
                 r2cs <- 1 - exp(-(nullDev - dev) / n)
                 r2n <- r2cs / (1 - exp(-nullDev / n))
+                r2t <- unname(diff(tapply(predict(self$models[[i]], type="response"),
+                                          self$models[[i]]$y, mean, na.rm=TRUE)))
 
-                pR2[[i]] <- list(r2mf=r2mf, r2cs=r2cs, r2n=r2n)
+                pR2[[i]] <- list(r2mf=r2mf, r2cs=r2cs, r2n=r2n, r2t=r2t)
             }
             return(pR2)
         },
@@ -706,6 +708,7 @@ logRegBinClass <- if (requireNamespace('jmvcore')) R6::R6Class(
             }
 
             predictTitle <- .("Predicted")
+            grmembshipTitle <- .("Group Membership")
             residsTitle <- .("Residuals")
             cooksTitle <- .("Cook's distance")
 
@@ -725,6 +728,7 @@ logRegBinClass <- if (requireNamespace('jmvcore')) R6::R6Class(
             if (self$nModels == 1) {
 
                 self$results$predictOV$set(1, title(predictTitle, dep), description(predictDescPrefix), 'continuous')
+                self$results$grmembershipOV$set(1, grmembshipTitle, description(grmembshipTitle), 'continuous')
                 self$results$residsOV$set(1, residsTitle, description(residsTitle), 'continuous')
                 self$results$cooksOV$set(1, cooksTitle, description(cooksTitle), 'continuous')
 
@@ -736,6 +740,10 @@ logRegBinClass <- if (requireNamespace('jmvcore')) R6::R6Class(
                 titles <- vapply(keys, function(key) title(predictTitle, paste(dep, key)), '')
                 descriptions <- vapply(keys, function(key) description(predictDescPrefix, key), '')
                 self$results$predictOV$set(keys, titles, descriptions, measureTypes)
+
+                titles <- vapply(keys, function(key) title(grmembshipTitle, key), '')
+                descriptions <- vapply(keys, function(key) description(grmembshipTitle, key), '')
+                self$results$grmembershipOV$set(keys, titles, descriptions, measureTypes)
 
                 titles <- vapply(keys, function(key) title(residsTitle, key), '')
                 descriptions <- vapply(keys, function(key) description(residsTitle, key), '')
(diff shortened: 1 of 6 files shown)


R/logregbin.b.R
```diff
@@ -1011,6 +1020,12 @@ logRegBinClass <- if (requireNamespace('jmvcore')) R6::R6Class(
                     self$results$predictOV$setValues(index=i, as.numeric(self$predicted[[i]]))
             }
 
+            if (self$options$grmembershipOV && self$results$grmembershipOV$isNotFilled()) {
+                self$results$grmembershipOV$setRowNums(private$.getDataRowNums())
+                for (i in seq_along(self$fitted))
+                    self$results$grmembershipOV$setValues(index=i, ifelse(self$fitted[[i]] > 0.5, 1, 0))
```
raviselker: I'm not a big fan of adding the group membership as an output variable like this as it is based on a fixed, arbitrary cutoff criterion of 0.5. At the moment it's easy to calculate the group membership in the spreadsheet by transforming the fitted values, which you can already get as an output variable. So I'd either just leave it out completely, or let the user choose a cutoff value to calculate group membership.

Nonetheless, I think if you want to do the latter you should make this a separate commit (and Pull Request). In general, one pull request should contain one piece of functionality.

R/logregbin.b.R
```diff
@@ -288,8 +288,10 @@ logRegBinClass <- if (requireNamespace('jmvcore')) R6::R6Class(
                 r2mf <- 1 - dev/nullDev
                 r2cs <- 1 - exp(-(nullDev - dev) / n)
                 r2n <- r2cs / (1 - exp(-nullDev / n))
+                r2t <- unname(diff(tapply(predict(self$models[[i]], type="response"),
+                                          self$models[[i]]$y, mean, na.rm=TRUE)))
```
raviselker: Maybe this would be a little bit more descriptive if it was calculated in two steps (with some added formatting):
```suggestion
                meanFittedProbs <- tapply(
                    self$models[[i]]$fitted.values, self$models[[i]]$y, mean, na.rm=TRUE
                )
                r2t <- unname(diff(meanFittedProbs))
```

jamovi/logregbin.u.yaml
```diff
@@ -165,6 +165,10 @@ children:
                     type: CheckBox
                     optionName: pseudoR2
                     optionPart: r2n
+                  - name: pseudoR2_r2t
```
raviselker, line 168: A general comment: could you add a test for this new statistic. You can just add it to the general "All options in the logRegBin work (sunny)" test in the `tests/testthat/testlogregbin.R` test file

raviselker (commented):
Thank you for the PR! So as I also said in my review, can you take the "group membership" functionality out of the commit so that this PR only focusses on the added Tjur's R2? I added a couple of comments for this as well, but I'm happy to merge this if these are addressed.

R/logregbin.b.R
MAgojam: You're right, two steps make it more readable

jamovi/logregbin.u.yaml
MAgojam: I will do it.

R/logregbin.b.R
MAgojam: First of all thank you for your time.

> I'm not a big fan of adding the group membership as an output variable like this as it is based on a fixed, arbitrary cutoff criterion of 0.5.  At the moment it's easy to calculate the group membership in the spreadsheet ... 

I can agree with you, but my observation says that there is a demand for this (SPSS users) and the ability to put them on or off with a simple click makes for a better user experience than having to figure them out.

> Nonetheless, I think if you want to do the latter you should make this a separate commit (and Pull Request).

As you suggest, it would be more interesting to accompany the option, adding the possibility of inserting a free cut-off.
If you feel I can make a pass with a PR just for that.

raviselker: Yeah, I think it would be a useful addition with the cutoff criterion added so if you turn this into a separate pull request, I'm sure we can get it to a place where we can merge it

R/logregbin.b.R
raviselker: I still see this as a oneliner in the code at the moment. Did you decide this was not a good idea anyway or did something go wrong with the commits?

raviselker:
I still see a bunch of commented out code. Could you please remove this? And I'm not sure if you're familiar with squashing your commits (see for instance, https://www.git-tower.com/learn/git/faq/git-squash) but it would be great if you could squash all your commits into one commit.

jonathon-love:
> it would be great if you could squash all your commits into one commit.

i'd leave squashing to me when i merge ... its better if we see the changes to what's come before, and squashing discards all that.

R/logregbin.h.R
```diff
@@ -230,6 +231,8 @@ logRegBinOptions <- if (requireNamespace("jmvcore", quietly=TRUE)) R6::R6Class(
                 hidden=TRUE)
             private$..predictOV <- jmvcore::OptionOutput$new(
                 "predictOV")
+            private$..grmembershipOV <- jmvcore::OptionOutput$new(
+                "grmembershipOV")
```
raviselker: Your generated analysis file still contains the group membership stuff. Could you rebuild the analysis (`jmvtools::install()`) and commit the changes in this file?

MAgojam: Sorry, gross oversight.
Unfortunately I was away from my PC and interacted from an iPad.
I'll do it before the sun comes up in the east.

raviselker (approved):
Looks good now, @jonathon-love could you clean up the commit history before merging it?

PR #385, opened 2023-05-26
contTables: Fisher's exact test can use a Monte Carlo simulation if the default work area is insufficient

MAgojam:
In contTables, if Fisher's exact test returns an error for an insufficient predefined work area, it is retried by activating with a Monte Carlo simulation to determine the p-value.


R/conttables.b.R
```diff
@@ -492,6 +499,9 @@ contTablesClass <- R6::R6Class(
                 if (inherits(fish, 'htest') && all(dim(mat) == 2) && hypothesis != "different")
                     chiSq$addFootnote(rowNo=othRowNo, 'p[fisher]', hypothesisTested)
 
+                if (MCpsimul)
+                    chiSq$addFootnote(rowNo=othRowNo, 'p[fisher]', .('Monte Carlo simulation (B=2000)'))
```
jonathon-love: shift the ` (B=2000)` outside the translation string

R/conttables.b.R
```diff
@@ -342,6 +342,14 @@ contTablesClass <- R6::R6Class(
                     dp <- NULL
                     lor <- NULL
                     fish <- try(stats::fisher.test(mat, conf.level=ciWidth, alternative=Ha), silent=TRUE)
+
+                    if (base::inherits(fish, 'try-error')) {
+                        # Monte Carlo simulation for p-value
+                        fish <- try(stats::fisher.test(mat, alternative=Ha, simulate.p.value=TRUE), silent=TRUE)
```
jonathon-love, line 348: do you need to specify the `conf.level` here? like the previous call?

R/conttables.h.R
```diff
@@ -33,14 +33,16 @@ contTablesOptions <- if (requireNamespace("jmvcore", quietly=TRUE)) R6::R6Class(
...
+            resS = FALSE,
             barplot = FALSE,
             yaxis = "ycounts",
             yaxisPc = "total_pc",
             xaxis = "xrows",
             bartype = "dodge", ...) {
 
             super$initialize(
-                package="jmv",
+                package="magojamtests",
```
jonathon-love: !!

R/conttables.h.R
```diff
@@ -701,13 +729,13 @@ contTablesBase <- if (requireNamespace("jmvcore", quietly=TRUE)) R6::R6Class(
                 requiresMissings = FALSE)
         }))
 
-#' Contingency Tables
+#' Tabella di Contingenza
```
jonathon-love: !!

R/conttables.h.R
```diff
@@ -340,6 +357,17 @@ contTablesResults <- if (requireNamespace("jmvcore", quietly=TRUE)) R6::R6Class(
...
+                options=options,
+                name="phoc",
+                title="Post-hoc (residuals)",
+                visible="(resP || resS)",
+                columns=list(),
+                clearWith=list(
+                    "rows",
+                    "cols",
+                    "counts",
+                    "layers")))
```
jonathon-love: !!

jonathon-love (changes requested):
hey mz,

sorry for the delay ... i think you've got some stray content in your .a.yaml file which is showing up in the .h.R file.

otherwise this is looking good.

MAgojam:
> sorry for the delay ... i think you've got some stray content in your .a.yaml file which is showing up in the .h.R file.

Sorry, you're right, I hadn't noticed.
:sob: It won't happen again!
Unfortunately I passed (stupidly) the two conttables files .b.r and .h.r of a module that I use as a clone (so as not to use jmv) in which I am adding a Post-hoc for a significant chi-square with the Standardized and Adjusted residuals (for an upcoming listing proposal for a contingency table) .

R/conttables.b.R
MAgojam: We can also eliminate `(B=2000)`, considering that it is the default value.

R/conttables.b.R
MAgojam: I can put it back (as in the above) even if it is ignored in the Monte Carlo simulation.

R/conttables.h.R
MAgojam: Last reply, below.

R/conttables.h.R
MAgojam: Last reply, below.

R/conttables.b.R
jonathon-love: yeah, lets eliminate for now.

R/conttables.b.R
jonathon-love: oh ok, if it's not used, then i agree it isn't necessary

jonathon-love:
merged. thanks mz!

PR #395, opened 2023-10-04
Fixing typos and creating a WORDLIST

sjentsch:
It's mainly fixing typo's and creating a WORDLIST (used by `devtools::spell_check()`). Australian English wasn't available, hence the `en-GB` in `DESCRIPTION`.
I wasn't able to build this version, but I checked `devtools::run_examples(run_dontrun = TRUE)` and `devtools::test()` which both run through.


NAMESPACE
```diff
@@ -28,8 +28,35 @@ export(ttestOneS)
...
+importFrom(jmvcore,glm)
+importFrom(jmvcore,jmv)
+importFrom(jmvcore,lm)
 importFrom(jmvcore,matchSet)
+importFrom(jmvcore,names)
+importFrom(jmvcore,of)
+importFrom(jmvcore,table)
+importFrom(jmvcore,the)
+importFrom(jmvcore,they)
+importFrom(jmvcore,warnings)
```
jonathon-love: this is all super-helpful, but i assume you can see this NAMESPACE just isn't right ... the unit-tests correctly point out that none of these functions occur in `jmvcore`.

if you wanted to push this forward, and it would be really helpful if you did, you could investigate what's going on here. i think i've seen this before, and i think the way i solve it is to use roxygen2 6.1.1 ...

jonathon-love (changes requested):

NAMESPACE
sjentsch: Are you aware of any method that enforces a particular version of Roxygen? Whenever I run commands from devtools, it seems to forcibly change the Roxygen-version to 7.2.3.

jonathon-love: you can get the 6.1.1 [here](https://cran.r-project.org/src/contrib/Archive/roxygen2/roxygen2_6.1.1.tar.gz) and install it with `R CMD install roxygen2_6.1.1.tar.gz` or similar.

PR #409, opened 2024-06-25
anovaRM: use full model if terms are not defined

### LINKED ISSUE #397: RM ANOVA ignores between subjects terms when called from R
details here:

https://forum.jamovi.org/viewtopic.php?p=11464#p11464

raviselker:
Previously, the rmTerms and the bsTerms always had to be defined
by the user of `jmv`. This commit changes it so that if you
don't define or or both terms, the model terms will default
to the full model including all possible terms.

Other than that, I found that when running jmvtools::install(), 
and running the rm anova within jamovi, jamovi complains that the
added packages are not available. Not sure whether this is a 
"real" issue, but I also added a commit that fixes this (feel
free to not merge this commit, if it's not).

Fixes #397

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -41,7 +41,12 @@ Imports:
     MASS,
     ggrepel,
     dplyr,
-    magrittr
+    magrittr,
+    reshape2,
+    lme4,
+    lmerTest,
+    estimability,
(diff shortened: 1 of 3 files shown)


jonathon-love:
merged

PR #410, opened 2024-07-02
Improve reference level handling

raviselker:
To improve the reference level handling in the analyses that make use of them, I defaulted the reference level to the first level of a factor if no or wrong reference level is defined.

The analyses that make use of the reference level updater are:
 - linReg
 - logRegBin
 - logRegMulti
 - logRegOrd
 - logLinear

### DIFF
--- a/R/linreg.b.R
+++ b/R/linreg.b.R
@@ -114,8 +114,21 @@ linRegClass <- R6::R6Class(
             return(private$.emMeans)
         },
         refLevels = function() {
-            if (is.null(private$.refLevels))
-                private$.refLevels <- private$.getRefLevels()
+            if (is.null(private$.refLevels)) {
+                refLevels <- getReferenceLevels(
+                    self$data, self$options$factors, self$options$refLevels
+                )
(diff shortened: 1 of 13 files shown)


R/linreg.b.R
```diff
@@ -112,6 +112,19 @@ linRegClass <- R6::R6Class(
...
+        },
+        refLevels = function() {
+            if (is.null(private$.refLevels)) {
+                refLevels <- getReferenceLevels(
+                    self$data, self$options$factors, self$options$refLevels
+                )
+                private$.refLevels <- refLevels$refLevels
+
+                if (length(refLevels$changedVars) > 0)
+                    setRefLevelWarning(self, refLevels$changedVars)
```
jonathon-love, line 124: side effects like this are *very* frowned upon!

raviselker: So how would you do this? Move it to the init? I would do this if you hadn't already merged this PR. Gotta give me a chance to do something with your comments :-P

PR #416, opened 2024-10-26
Added Mahalanobis distance to linear regression

sjentsch:
Added Mahalanobis distance (both as a results table and as output data) to linear regression


R/linreg.b.R
```diff
@@ -654,6 +683,13 @@ linRegClass <- R6::R6Class(
                 descriptions <- vapply(keys, function(key) description(.("Cook's distance"), key), '')
                 self$results$cooksOV$set(keys, titles, descriptions, measureTypes)
             }
+
+            titles <- vapply(c("ChiSq", "p"), function(name) title(.("Mahalanobis"), name), '')
+            descriptions <- vapply(c("Chi Sq.", "p"), function(s)
+                                   jmvcore::format(.("Mahalanobis distance ({stat}) for the variables in \"Covariates\""), stat = s),
+                                   ''
+                                  )
+            self$results$mahalOV$set(seq(2), titles, descriptions, rep('continuous', 2))
```
raviselker: At the moment, the Mahalanobis distance does not support adding variables in multiple blocks. If you add two variables in `Block 1`, and one variable in `Block 2` under the Model Builder, only one set of Mahalanobis distance scores is saved to the spreadsheet. However, each model should have its own associated Mahalanobis distance scores (similar to Cooks distance).

R/linreg.b.R
```diff
@@ -850,6 +886,37 @@ linRegClass <- R6::R6Class(
...
+                return()
+
+            mahal <- self$mahal
+            mahalp <- as.numeric(self$options$mahalp)
+            groups <- self$results$models
+            termsAll <- private$.getModelTerms()
+
+            for (i in seq_along(termsAll)) {
+                table <- groups$get(key=i)$dataSummary$mahal
+                if (is.null(mahal[[i]])) {
```
raviselker: When I only add one variable as a covariate I get the following error:
```
Error in mahal[[i]]: subscript out of bounds

private$.run()
```

R/linreg.b.R
```diff
@@ -850,6 +886,37 @@ linRegClass <- R6::R6Class(
...
+                    table$setNote("row", .("Mahalanobis distance can only be calculated for models with two or more dependent variables."))
```
raviselker: It would be nicer if the user gets a `Notice` (see utils `setAnalysisNotice`) function when the Mahalanobis distance cannot be calculated.

R/linreg.b.R
```diff
@@ -850,6 +886,37 @@ linRegClass <- R6::R6Class(
...
+                } else {
+                    # select values below p-threshold (outliers)
+                    selRow <- which(mahal[[i]]$p <= mahalp)
+                    for (j in selRow)
+                        table$addRow(rowKey = j, values = mahal[[i]][j, ])
+                    table$setNote(
+                        "row",
+                        jmvcore::format(.("Filter syntax: {filter}"),
+                            filter = paste(paste0("ROW() != ", mahal[[i]][selRow, "row"]), collapse = " and "))
+                    )
```
raviselker: I'm not a fan of this note because:
1. I think it would be easier (and more robust) to use the saved Mahalanobis distance to filter out rows
2. This note only makes sense if you use jamovi, and not if you're using `jmv` through R directly

R/linreg.b.R
```diff
@@ -285,6 +293,27 @@ linRegClass <- R6::R6Class(
...
+                    return(NULL)
+                }
+            }
+
+            mahal <- list()
+            data <- private$.cleanData(naSkip=jmvcore::toB64(self$options$dep))
+            for (i in seq_along(self$models))
+                mahal[[i]] <- mahal2List(data, attr(self$models[[i]]$terms, "term.labels"))
+            
+            mahal[[i + 1]] <- mahal2List(data, jmvcore::toB64(self$options$dep))
```
raviselker: Why do you do this? Wouldn't this always retun `NULL`?

R/linreg.b.R
```diff
@@ -850,6 +886,37 @@ linRegClass <- R6::R6Class(
                 table$setRow(rowNo=1, values=cooks[[i]])
             }
         },
+        .populateMahalTable = function() {
```
raviselker, line 891: This table now works differently from the cooks table, which is more of a summary of the distances. The nice thing about a summary table is that you get a quick overview of the distribution of values, and if you want to do something based on these values it makes more sense to use the saved mahalanobis distances. So I'd prefer a similar table to the Cook's distance would be given here.

R/linreg.b.R
```diff
@@ -1020,6 +1087,13 @@ linRegClass <- R6::R6Class(
                 for (i in seq_along(self$cooks))
                     self$results$cooksOV$setValues(index=i, self$cooks[[i]])
             }
+
+            if (self$options$mahalOV && self$results$mahalOV$isNotFilled()) {
```
raviselker, line 1088: This only works for a single model, not when you use the model builder and add multiple blocks

tests/testthat/testlinreg.R
raviselker, line 1: Could you exclude file mode changes from the git commit (i.e., by `git config core.fileMode false`)

I think each added feature should test it. I think the tests should also represent different use cases, such as multi block analyses, and fewer than 2 dependent variables to make sure these use cases are done correctly

tests/testthat/testmancova.R
raviselker, line 1: Could you exclude file mode changes from the git commit (i.e., by git config core.fileMode false)

raviselker (changes requested):
Hi @sjentsch, I think adding the mahalanobis distance measures would be a good idea. However, I think you'd need to address a couple of issues for it to be merged. If you need some help, lmk through slack

R/linreg.b.R
jonathon-love: > I think it would be easier (and more robust) to use the saved Mahalanobis distance to filter out rows

i think this is necessary because you can't filter on output variables ... it creates a circular dependency ... you apply a filter which changes the analysis, which changes the output column, which changes the filter, which changes the analysis ... and so on.

tests/testthat/testmancova.R
jonathon-love: these changes are in a separate commit, so that seems OK to me?

R/linreg.b.R
sjentsch: I will look into that. I thought I had a if-condition that handles that.

R/linreg.b.R
sjentsch: Comment (1) on Slack.

R/linreg.b.R
sjentsch: I agree and will change that. What you propose makes more sense.

R/linreg.b.R
sjentsch: Point 2, I can't really solve (but people using `jmv` on R may be knowledgable enough to “translate” jamovi-filter into something that can be used in R.
The comment of jonathon on point 1 is exactly my point (you can't use the output variables for filtering).

sjentsch: But I am certainly will listen if there is a better way of solving this.

R/linreg.b.R
sjentsch: This may go away if we create the output variables model-wise. It is currently there to provide the output variables for all “Covariates” in the model.

tests/testthat/testlinreg.R
sjentsch: Yes, and I should perhaps not have changed them. I just thought these are R-files and the should not be executable. Just drop the commit, if you disagree.

R/linreg.b.R
sjentsch: Fixed.

R/linreg.b.R
sjentsch: Fixed.

R/linreg.b.R
sjentsch: I tried, however, I don't think it improves things. Since the Mahalanobis distance can only be calculated with two or more variables, one would see this error message every time one runs a hierarchical regression where the first block contains only one variable. This might be annoying for the users?

sjentsch: It would make sense if I could attach it to the group / the block. If this is possible, let me know, I could not figure out how.

R/linreg.b.R
sjentsch: Resolved because I now calculate Mahalanobis distances for each block.

R/linreg.b.R
sjentsch: Fixed.

R/linreg.b.R
jonathon-love: longer term, we're going to support multiple "sheets" within a "project", and one would be able to make a second "sheet" containing the data filtered by m9s so this filter "formula output" won't be necessary then, but we'll have to keep it in because people don't cope when you remove things. ... and of course if we keep it in, people will assume that's the "recommended way" to do it.

how about we approach this a different way @sjentsch, what if you instruct people to copy the m9s output variable into a new data variable, and then filter on that?

i appreciate it's hackish and ugly ... but it might be the lesser of several evils.

thoughts?

R/linreg.b.R
raviselker: Yeah, that's indeed how I would do it. You can manually add a notice by doing the following (I tested it, and it works):
```R
                    notice <- jmvcore::Notice$new(
                        options = self$options,
                        name = 'warningMessage',
                        type = jmvcore::NoticeType$STRONG_WARNING
                    )
                    notice$setContent(.("Mahalanobis distance can only be calculated for models with two or more independent variables."))
                    groups$get(key=i)$insert(1, notice)
```

R/linreg.b.R
raviselker: Yeah, I'd be fine with this workaround

tests/testthat/testmancova.R
raviselker: Yeah, but this feature shouldn't have to touch the `testmancova` file, so it's a bit strange that it's changed in this MR. For the other file (the `testlinreg` I'm fine with keeping it)

R/linreg.b.R
```diff
@@ -285,6 +293,21 @@ linRegClass <- R6::R6Class(
 
             return(cooksSummary)
         },
+        .computeMahal = function() {
+            mahal <- vector(mode = "list", length(self$nModels))
+            data <- private$.cleanData(naSkip=jmvcore::toB64(self$options$dep))
```
raviselker: Is there a reason you don't just use the `dataProcessed` active binding?

R/linreg.b.R
```diff
@@ -285,6 +293,21 @@ linRegClass <- R6::R6Class(
...
+            for (i in seq_along(self$models)) {
+                cov <- attr(self$models[[i]]$terms, "term.labels")
+                cov <- intersect(cov, jmvcore::toB64(self$options$covs))
```
raviselker, line 304: Are factors not relevant for calculating the mahalanobis distance? At this moment it's calculated only based on the covariates.

R/linreg.b.R
sjentsch: Using a notice is implemented now (see above).

R/linreg.b.R
sjentsch: I removed the filter syntax. This will enable people to either copy the output columns and use that or to just create a filter syntax by writing the line numbers with the outliers manually (like I did).
@raviselker: I asume the line numbers match those of the dataset when I use `self$data`?

jonathon-love: if you've got filtered rows, then these won't appear in `self$data` ... so if the first row was filtered out of the spreadsheet, the second row in the spreadsheet would appear in the first row in `self$data`, the third row in the spreadsheet would appear in the second row of `self$data`, etc.

you can get the actual row numbers with `row.names(self$data)`

sjentsch: Yes, but I actually do this. What is in the column `rows`  gets assigned `return(data.frame(row = row.names(data), chisq = mahalChiSq, p = pchisq(mahalChiSq, ndep, lower.tail = FALSE)))` (l. 302) with the data coming from `.cleanData()`  which assigns the correct `row.names` (l. 1456). It's perhaps not obvious because of the three steps, but in large parts similar to what Ravi (or whoever wrote these parts of `linreg.b.R`) does.

raviselker: So the `private$.getDataRowNums()` gives the row numbers of the `self$dataProcessed` data set. However, for some reason you use `private$.cleanData(naSkip=jmvcore::toB64(self$options$dep))` instead. Therefore, it's possible that the length of the `getDataRowNums` and the cleaned data do not match, which would result (I assume) in an error.

So a solution would be etiher one of these:
1. use `self$dataProcessed` instead (just like I've done in the cooks distance)
2. get the rownames from the mahal distance itself

Unless there's a good reason to not use `self$dataProcessed`, I'd go for option `1`

R/linreg.b.R
sjentsch: No, it is only for continuous variables.

R/linreg.b.R
sjentsch: Oh, I did not know. I just used it because you use it a couple of rows up in `.computePredicted`.

raviselker (approved):
I checked the changes, and they address the issues I originally had so this can me merged

raviselker:
Because there were some conflicts with this branch, I created my own branch based on this branch and fixed the issues. I opened a new PR for this branch (https://github.com/jamovi/jmv/pull/421) so this one can be closed.

PR #417, opened 2024-10-29
Add "styler" autoformatter configuration

raviselker:
This commit configures the styler autoformatter and makes it easy to use over the whole project by adding a couple of Makefile commands.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -48,6 +48,7 @@ Suggests:
     semPlot,
     carData,
     knitr,
-    rmarkdown
+    rmarkdown,
+    styler
 Encoding: UTF-8
 RoxygenNote: 6.1.1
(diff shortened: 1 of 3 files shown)


DESCRIPTION
```diff
@@ -48,6 +48,7 @@ Suggests:
     semPlot,
     carData,
     knitr,
-    rmarkdown
+    rmarkdown,
+    styler
```
jonathon-love, line 52: i think we don't need to include this in suggests?

it will end up shipping with jamovi, which we don't need.

raviselker:
Looks like there's a better alternative for autoformatting (https://www.tidyverse.org/blog/2025/02/air/), so I'd rather look into this one. I'll close this PR.

PR #419, opened 2024-12-03
Added post-hoc table for residuals

MAgojam:
## Summary of Changes

This PR introduces a dedicated **post-hoc residuals table (phoc)** to complement the contingency table analyses.
Alongside the primary frequency tables and their associated statistics, the phoc table provides a richer diagnostic layer by presenting an array of residuals at the cell level.

### Key Features:

* **New phoc table:** Adds a post-hoc overview of cell-level residuals (Unstandardized, Pearson, Standardized, and Adjusted).
* **Residual Types:**
  * **Unstandardized:** Raw differences between observed and expected counts.
  * **Pearson:** Residuals scaled by the square root of the expected values.
  * **Standardized:** Residuals further normalized to account for row and column distributions, facilitating clearer interpretation.
  * **Adjusted:** Residuals derived from a Poisson GLM, providing an additional layer of refinement by incorporating row and column effects.
* **Integration:** Computations are seamlessly integrated into the existing workflow. A chi-square test and a GLM fit are performed internally, and their results are then rendered in the phoc table.
* **Enhanced Testing:** The `testconttables.R` file has been updated with new tests, ensuring the accuracy and consistency of each residual type.

### User Benefit:

This enhancement offers a deeper analytical perspective, enabling users to pinpoint meaningful discrepancies and interpret their contingency table results with greater clarity and nuance.

### DIFF
--- a/R/conttables.b.R
+++ b/R/conttables.b.R
@@ -246,6 +246,8 @@ contTablesClass <- R6::R6Class(
             gamma$getColumn('ciu')$setSuperTitle(ciText)
 
             private$.initBarPlot()
+
+            private$.initPhocTab(data=data)
         },
         .run=function() {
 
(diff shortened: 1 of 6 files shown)


jamovi/conttables.u.yaml
```diff
@@ -194,6 +194,27 @@ children:
                         name: pcCol
                       - type: CheckBox
                         name: pcTot
+              - type: LayoutBox
```
raviselker, line 206: I think it would be nice if the post hoc tests are in its own CollapseBox so that it's in line with the post hoc tests in the anova.

jamovi/conttables.u.yaml
```diff
@@ -194,6 +194,27 @@ children:
...
+                cell:
+                  column: 2
+                  row: 0
+                stretchFactor: 1
+                children:
+                  - type: Label
+                    label: Post-hoc
```
raviselker: We do labels in title case. Also, try to keep labels consistent with other analyses. In the anova it's called `Post Hoc Tests`.

jamovi/conttables.r.yaml
```diff
@@ -365,4 +365,21 @@ items:
         - yaxis
         - yaxisPc
         - xaxis
-        - bartype
\ No newline at end of file
+        - bartype
+
+    - name: phoc
```
raviselker: The post hoc table in anova is called `postHoc` (see https://github.com/jamovi/jmv/blob/9643e88c4eb8a205001b448f1051d4acf0dac619/jamovi/ancova.r.yaml#L165), please change it so that it's consistent with that

jamovi/conttables.r.yaml
```diff
@@ -365,4 +365,21 @@ items:
...
+      title: Post-hoc (residuals)
```
raviselker: I would just call it `Post Hoc Tests` (you can see that it's residuals from the table)
```suggestion
      title: Post Hoc Tests
```

jamovi/conttables.r.yaml
```diff
@@ -365,4 +365,21 @@ items:
...
+
+      clearWith:
+        - rows
+        - cols
+        - counts
+        - layers
+        - resU
+        - resP
+        - resS
+        - resA
```
raviselker: You probably don't need to clear the table when these options change as fitted model should stay the same

jamovi/conttables.r.yaml
```diff
@@ -365,4 +365,21 @@ items:
...
\ No newline at end of file
+        - bartype
+
+    - name: phoc
+      title: Post-hoc (residuals)
+      type: Table
+      visible: (resU || resP || resS || resA)
+      description: a table of residuals
+      columns: [ ]
+
```
raviselker:
```suggestion
```

jamovi/conttables.r.yaml
```diff
@@ -365,4 +365,21 @@ items:
...
         - xaxis
-        - bartype
\ No newline at end of file
+        - bartype
+
+    - name: phoc
+      title: Post-hoc (residuals)
+      type: Table
+      visible: (resU || resP || resS || resA)
+      description: a table of residuals
```
raviselker:
```suggestion
      description: a table of post-hoc residuals
```

jamovi/conttables.r.yaml
raviselker: Can you move this table above the `barPlot`? I think it's nicer if the plots are shown at the bottom of the results.

R/conttables.b.R
```diff
@@ -590,8 +592,288 @@ contTablesClass <- R6::R6Class(
...
+                key <- paste0(rows[i,], collapse='`')
+                phoc$addRow(rowKey=key, values=values)
+
+                if (nextIsNewGroup) {
+                    phoc$addFormat(rowNo=i, 1, Cell.BEGIN_GROUP)
+                    nextIsNewGroup <- FALSE
+                }
+            }
         },
+        .runPhocTab=function(data=data) {
```
raviselker: Can you call it `.populatePostHocTable` to make it consistent with other analyses? See for instance the linear regression (https://github.com/jamovi/jmv/blob/master/R/linreg.b.R) for an example on naming conventions.

R/conttables.b.R
```diff
@@ -590,8 +592,288 @@ contTablesClass <- R6::R6Class(
...
+                name=title,
+                title=title,
+                type='text')
+
+            # add the column columns (from the column variable)
+            # fill in dots, if no column variable specified
+
+            if ( ! is.null(colVarName)) {
+                superTitle <- colVarName
+                levels <- base::levels(data[[colVarName]])
```
raviselker: You don't have call functions from the `base` namespace explicitly (otherwise you'd have to `base::` every base function that you use. So this can just be:
 ```suggestion
                 levels <- levels(data[[colVarName]])
```

File: R/conttables.b.R [hunk 10]
```diff
@@ -590,8 +592,288 @@ contTablesClass <- R6::R6Class(
...
+                levels <- c('.', '.')
+            }
+
+            # iterate over the sub rows
+            for (j in seq_along(subNamesPh)) {
+                subName <- subNamesPh[[j]]
+                if (subName == '[resU]')
+                    vPh <- '(resU && (resP || resS || resA))'
+                else
+                    vPh <- visiblePh[j]
```
raviselker: So now the `residuals` column is not visible if **only** the unstandardized residuals are checked. I think it's nice to not show the column when only one type of `residuals` is checked no matter what the type is. However, than you do need to add some information (probably to the table title) that you are showing a specific type of post hoc test. So for instance, for the unstandardized residuals I would expect a title such as `Post Hoc Tests (Unstandardized Residuals).

R/conttables.b.R
```diff
@@ -590,8 +592,288 @@ contTablesClass <- R6::R6Class(
...
+                    # Formatting for significant residues
+                    if (self$options$resP) {
+                        for (colIndex in seq_len(nCols)) {
+                            colName <- paste0(colIndex, '[resP]')
+                            resValue <- residualsP[rowNo, colIndex]
+                            if (!is.na(resValue) && abs(resValue) > 2) {
+                                phoc$addFormat(rowNo=freqRowNo, col=colName, Cell.NEGATIVE)
+                                phoc$addFootnote(rowNo=freqRowNo,
+                                                 col=colName,
+                                                 .('Pearson Residuals: (Observed - Expected) / sqrt(Expected).'))
```
raviselker: I don't think these footnotes add much as it's very easy to look up what each specific residual means. Moreover, these footnotes would be more general information that wouldn't only show up when a value is `> 2`.

R/conttables.b.R
```diff
@@ -590,8 +592,288 @@ contTablesClass <- R6::R6Class(
...
+
+                    phoc$setRow(rowNo=freqRowNo, values=values)
+
+                    # Formatting for significant residues
+                    if (self$options$resP) {
+                        for (colIndex in seq_len(nCols)) {
+                            colName <- paste0(colIndex, '[resP]')
+                            resValue <- residualsP[rowNo, colIndex]
+                            if (!is.na(resValue) && abs(resValue) > 2) {
+                                phoc$addFormat(rowNo=freqRowNo, col=colName, Cell.NEGATIVE)
```
raviselker: At the moment, it's not really clear what it means if a value turns red. I think it would be good to add a general table note if any of the values become red so that users know what this indicates.

Maybe it would also be nice to make the highlighting of values an option (similar to how it's done in CFA):
Screenshot 2024-12-04 at 12 53 40

tests/testthat/testconttables.R
```diff
@@ -62,6 +66,21 @@ testthat::test_that('All options in the contTables work (sunny)', {
     testthat::expect_equal(c(0.45, 0.55, 1), mainTable[['.total[pcCol]']], tolerance = 1e-3)
     testthat::expect_equal(c(0.45, 0.55, 1), mainTable[['.total[pcTot]']], tolerance = 1e-3)
 
+    # Test residuals phoc tables
```
raviselker: Appreciate the addition of test cases! Are there any alternative scenarios that should be tested?

raviselker (changes requested):
Hi @MAgojam,
Awesome work, i think this will make a great addition to the contingency table analysis! Overall, I think this PR is already in a pretty good state, but I've added some comments on where I think improvements can be made. Thanks for your work!
Cheers,
Ravi

tests/testthat/testconttables.R
MAgojam: Yes, I updated to also check for a layer variable

MAgojam:
Hi @raviselker,

Thank you for your feedback on my PR. I've carefully reviewed your comments and have made the necessary adjustments. I've also conducted additional testing to ensure that the changes work as intended.

Please let me know if you would like me to provide any further details.

Regards,
Maurizio

R/conttables.b.R
MAgojam: Now there is like you suggested

R/conttables.b.R
MAgojam: You're right, I deleted them.

R/conttables.b.R
MAgojam: I did as you suggested and you will tell me if it can work like this.

R/conttables.b.R
MAgojam: I did as you suggested, but since it was the same in the other functions that I did not touch and I had followed the same calling method.

jamovi/conttables.r.yaml
MAgojam: Yes you are right, so I removed the possibility of a `clearWith` for the `postHoc` table, because it would lead to an unnecessary step from the `.init()` function, where it would be rebuilt even when the values ​​of the residues have not been modified.

jamovi/conttables.r.yaml
MAgojam: Done

jamovi/conttables.u.yaml
MAgojam: Yes this is a great suggestion, which I have applied in this update.

jamovi/conttables.u.yaml
MAgojam: I know, I did it and I will be more careful about these aspects next time.

jamovi/conttables.r.yaml
MAgojam: Done.

R/conttables.b.R
MAgojam: Done.

jamovi/conttables.r.yaml
MAgojam: Done.

jamovi/conttables.r.yaml
MAgojam: Done.

jamovi/conttables.a.yaml
```diff
@@ -405,4 +405,63 @@ options:
...
             stack or side by side (default), barplot type
+
+    - name: resU
+      title: Unstandardized residuals
+      type: Bool
+      default: false
+      description:
+          R: >
+            `TRUE` or `FALSE` (default), provide Unstandardized residuals
+
```
raviselker, line 416: Is there a reason that it's not possible to highlight the unstandardized residuals?

raviselker (approved):
The rework looks good! I added one question, but that's my only comment

jamovi/conttables.a.yaml
MAgojam: The unstandardized residuals can technically be highlighted, but they are generally less informative in the context of contingency tables.
Unlike standardized residuals, unstandardized residuals are raw differences between observed and expected frequencies and do not account for the variability of the data or the sample size. Highlighting them may lead to misleading interpretations, as larger residuals could simply reflect larger expected frequencies rather than significant deviations.

If highlighting unstandardized residuals is critical for a specific analysis, it can be implemented, but care should be taken to clarify their limitations and ensure they are not misinterpreted. Alternatively, standardized residuals are typically used for highlighting because they normalize these differences, making them easier to compare across cells and interpret within the context of statistical significance.

I owe you one, raviselker!
Thanks for taking the time to review this PR.

jonathon-love: what does "highlight" mean in this context?

MAgojam: The `a.yaml `file currently defines highlighting thresholds only for three residue types: P, S, and A.
 This means that unstandardized residuals are not explicitly considered for highlighting.
 I previously explained the rationale for excluding unstandardized residuals.
 However, if you would like to include them for completeness, I can readily add the necessary configuration to the `a.yaml` file.

PR #425, opened 2025-02-12
descriptives: add weights support

raviselker:
Implements weights support for a number of descriptive statistics. Not all descriptives functionality is supported yet (most notable, the plots), but the user gets a warning message when they select an unsupported option.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -42,7 +42,8 @@ Imports:
     MASS,
     ggrepel,
     dplyr,
-    magrittr
+    magrittr,
+    matrixStats
 Suggests:
     exact2x2,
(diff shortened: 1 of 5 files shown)


jonathon-love:
did some testing; throws an error with quartiles

Screenshot 2025-02-23 at 14 40 03

[Tooth Growth 2.omv.zip](https://github.com/user-attachments/files/18927795/Tooth.Growth.2.omv.zip)

raviselker:
Good catch! For some reason I was unable to replicate it in my R version (maybe a different way of dealing with NA values in if statements between the jamovi R version and my local R version), but I was able to test it in jamovi using your example. I added a commit that fixes this.

PR #428, opened 2025-03-29
Post-Hoc: Dunn's test added

MAgojam:
Added Dunn's test without committing additional libraries (vanilla code).
Also updated for a check from testthat.

### DIFF
--- a/R/anovanp.h.R
+++ b/R/anovanp.h.R
@@ -8,11 +8,11 @@ anovaNPOptions <- if (requireNamespace("jmvcore", quietly=TRUE)) R6::R6Class(
         initialize = function(
             deps = NULL,
             group = NULL,
-            es = FALSE,
-            pairs = FALSE, ...) {
+            kwES = list(),
+            kwPH = list(), ...) {
 
             super$initialize(
-                package="jmv",
+                package="mzanovanp",
                 name="anovaNP",
                 requiresData=TRUE,
                 ...)
@@ -34,30 +34,35 @@ anovaNPOptions <- if (requireNamespace("jmvcore", quietly=TRUE)) R6::R6Class(
                     "ordinal"),
                 permitted=list(
                     "factor"))
-            private$..es <- jmvcore::OptionBool$new(
-                "es",
-                es,
-                default=FALSE)
-            private$..pairs <- jmvcore::OptionBool$new(
-                "pairs",
-                pairs,
-                default=FALSE)
+            private$..kwES <- jmvcore::OptionNMXList$new(
+                "kwES",
+                kwES,
+                options=list(
+                    "epsilonsq"),
+                default=list())
+            private$..kwPH <- jmvcore::OptionNMXList$new(
+                "kwPH",
+                kwPH,
+                options=list(
+                    "dscf",
+                    "dunn"),
+                default=list())
 
             self$.addOption(private$..deps)
             self$.addOption(private$..group)
-            self$.addOption(private$..es)
-            self$.addOption(private$..pairs)
+            self$.addOption(private$..kwES)
+            self$.addOption(private$..kwPH)
         }),
     active = list(
         deps = function() private$..deps$value,
         group = function() private$..group$value,
-        es = function() private$..es$value,
-        pairs = function() private$..pairs$value),
+        kwES = function() private$..kwES$value,
+        kwPH = function() private$..kwPH$value),
     private = list(
         ..deps = NA,
         ..group = NA,
-        ..es = NA,
-        ..pairs = NA)
+        ..kwES = NA,
+        ..kwPH = NA)
 )
 
 anovaNPResults <- if (requireNamespace("jmvcore", quietly=TRUE)) R6::R6Class(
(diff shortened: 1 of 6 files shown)


R/anovanp.h.R
```diff
@@ -8,11 +8,11 @@ anovaNPOptions <- if (requireNamespace("jmvcore", quietly=TRUE)) R6::R6Class(
...
             deps = NULL,
             group = NULL,
-            es = FALSE,
-            pairs = FALSE, ...) {
+            kwES = list(),
+            kwPH = list(), ...) {
 
             super$initialize(
-                package="jmv",
+                package="mzanovanp",
```
raviselker: This shouldn't have changed. If you reinstall the module, the `.h.R` files should be correct.

R/anovanp.h.R
```diff
@@ -130,20 +135,44 @@ anovaNPResults <- if (requireNamespace("jmvcore", quietly=TRUE)) R6::R6Class(
...
+                            `visible`="(kwPH:dunn)")))))}))
 
 anovaNPBase <- if (requireNamespace("jmvcore", quietly=TRUE)) R6::R6Class(
     "anovaNPBase",
     inherit = jmvcore::Analysis,
     public = list(
         initialize = function(options, data=NULL, datasetId="", analysisId="", revision=0) {
             super$initialize(
-                package = "jmv",
+                package = "mzanovanp",
```
raviselker: This shouldn't have changed. If you reinstall the module, the `.h.R` files should be correct.

jamovi/anovanp.a.yaml
```diff
@@ -76,19 +76,32 @@ options:
       description:
           R: a string naming the grouping or independent variable in `data`
 
-    - name: es
-      title: Effect size
-      type: Bool
-      default: false
```
raviselker: So the consequence of removing options is that you lose backwards compatibility with `.omv` files that were generated using an older version of `jmv`. So we try to avoid this if there's no very good reason to do so (and I don't think there is here). So could you please use the `es` option for the effect size instead of the newly created `kwES`.

jamovi/anovanp.a.yaml
```diff
@@ -76,19 +76,32 @@ options:
...
+          ui: >
+            provide a ε² measure of effect size for the Kruskal-Wallis tests.
           R: >
-            `TRUE` or `FALSE` (default), provide effect-sizes
+            this is epsilon squared, a type of effect size for the Kruskal-Wallis test
 
-    - name: pairs
-      title: DSCF pairwise comparisons
-      type: Bool
-      default: false
```
raviselker: Same as for the effect size, we don't really want to remove options if we can help. A simple way forward is by keeping the `pairs` option and let that one be used for the `dscf` comparisons, and creating a new boolean option for dunn.

tests/testthat/testanovanp.R
```diff
@@ -6,7 +6,7 @@ testthat::test_that('All options in the anovaNP work (sunny)', {
 
     data <- data.frame(f = factor, dep = dep)
 
-    r <- jmv::anovaNP(data, deps='dep', group='f', es=TRUE, pairs=TRUE)
```
raviselker: Because we want this analysis to be backwards compatible, I'd keep this test as it was, and create a new test to test the dunn's post-hoc tests.

jamovi/anovanp.r.yaml
```diff
@@ -34,19 +34,19 @@ items:
...
       type: Array
       description: an array of pairwise comparison tables
       items: (deps)
-      visible: (pairs)
+      visible: (kwPH:dscf || kwPH:dunn)
       clearWith:
         - group
 
       template:
-        title: Pairwise comparisons - $key
```
raviselker: I would just create to separate tables for dunn's and dscf, as they don't really share any statistics.

raviselker (changes requested):
Some things needs to be addressed (most importantly the backwards compatibility issue).

MAgojam:
Thank you @raviselker  for your time.
In this commit, the necessary corrections have been made, incorporating your suggestions and requests. Always available for anything that may be needed.

R/anovanp.h.R
MAgojam: Done

R/anovanp.h.R
MAgojam: DOne

jamovi/anovanp.a.yaml
MAgojam: Done

jamovi/anovanp.a.yaml
MAgojam: Done

jamovi/anovanp.r.yaml
MAgojam: Done

tests/testthat/testanovanp.R
MAgojam: Done.

jamovi/anovanp.u.yaml
```diff
@@ -20,32 +20,12 @@ children:
...
-          row: 0
-        children:
-          - name: kwPH_dscf
-            type: CheckBox
-            optionPart: dscf
-            optionName: kwPH
-          - name: kwPH_dunn
-            type: CheckBox
-            optionPart: dunn
-            optionName: kwPH
```
raviselker: So I did like that you put the effect size on the left and the pairwise comparisons on the right (both with a label). Could you add this again? So on the left a the effect size (you can change the title of the effect size option to ε², the title does not need to be the same in order to be backwards compatible), and on the right the Post-Hoc Tests / Pairwise Comparisons?

raviselker (changes requested):
Looking good @MAgojam! I just added one request for a small visual change to the options UI.

jamovi/anovanp.u.yaml
MAgojam: Done

raviselker (approved):
Looks good! This can be merged @jonathon-love

PR #434, opened 2025-09-09
Contrasts for the rmANOVA

sjentsch:
- implemented contrasts for the rmANOVA
- added simple_1, and simple_k (simple contrasts vs. either first vs. last category) to all ANOVA, ANCOVA (and rmANOVA)
- checked the results vs. SPSS (correct / same apart from rounding errors at ~ second decimal)

### DIFF
--- a/R/ancova.b.R
+++ b/R/ancova.b.R
@@ -188,7 +188,7 @@ ancovaClass <- R6::R6Class(
 
                 var <- data[[contrast$var]]
                 levels <- base::levels(var)
-                labels <- private$.contrastLabels(levels, contrast$type)
+                labels <- contrastLabels(levels, contrast$type) # defined in utilsanova.R
 
                 for (label in labels)
                     table$addRow(rowKey=label, list(contrast=label))
@@ -539,7 +539,7 @@ ancovaClass <- R6::R6Class(
                     next()
 
                 levels <- base::levels(data[[var]])
-                labels <- private$.contrastLabels(fromB64(levels), type)
+                labels <- contrastLabels(fromB64(levels), type) # defined in utilsanova.R
 
                 table <- self$results$get('contrasts')$get(contrast)
 
@@ -787,73 +787,11 @@ ancovaClass <- R6::R6Class(
         },
 
         #### Helper functions ----
-        .contrastLabels=function(levels, type) {
-
-            nLevels <- length(levels)
-            labels <- list()
-
-            if (length(levels) <= 1) {
-
-                # do nothing
-
-            } else if (type %in% c('simple_1', 'simple')) {
-
-                for (i in seq_len(nLevels-1))
-                    labels[[i]] <- paste(levels[i+1], '-', levels[1])
-
-            } else if (type == 'simple_k') {
-
-                for (i in seq_len(nLevels-1))
-                    labels[[i]] <- paste(levels[i], '-', levels[nLevels])
-
-            } else if (type == 'deviation') {
-
-                all <- paste(levels, collapse=', ')
-                for (i in seq_len(nLevels-1))
-                    labels[[i]] <- paste(levels[i+1], '-', all)
-
-            } else if (type == 'difference') {
-
-                for (i in seq_len(nLevels-1)) {
-                    rhs <- paste0(levels[1:i], collapse=', ')
-                    labels[[i]] <- paste(levels[i + 1], '-', rhs)
-                }
-
-            } else if (type == 'helmert') {
-
-                for (i in seq_len(nLevels-1)) {
-                    rhs <- paste(levels[(i+1):nLevels], collapse=', ')
-                    labels[[i]] <- paste(levels[i], '-', rhs)
-                }
-
-            } else if (type == 'repeated') {
-
-                for (i in seq_len(nLevels-1))
-                    labels[[i]] <- paste(levels[i], '-', levels[i+1])
-
-            } else if (type == 'polynomial') {
-
-                names <- c(
-                    .('linear'), .('quadratic'), .('cubic'), .('quartic'),
-                    .('quintic'), .('sextic'), .('septic'), .('octic')
-                )
-
-                for (i in seq_len(nLevels-1)) {
-                    if (i <= length(names)) {
-                        labels[[i]] <- names[i]
-                    } else {
-                        labels[[i]] <- paste('degree', i, 'polynomial')
-                    }
-                }
-            }
-
-            labels
-        },
         .createContrasts=function(levels, type) {
 
             nLevels <- length(levels)
 
-            if (type %in% c('simple_1', 'simple')) {
+            if (type == 'simple') {
 
                 dummy <- contr.treatment(levels, base = 1)
                 dimnames(dummy) <- NULL
--- a/R/ancova.b.R
+++ b/R/ancova.b.R
@@ -796,10 +796,15 @@ ancovaClass <- R6::R6Class(
 
                 # do nothing
 
-            } else if (type == 'simple') {
+            } else if (type %in% c('simple_1', 'simple')) {
 
                 for (i in seq_len(nLevels-1))
                     labels[[i]] <- paste(levels[i+1], '-', levels[1])
+
+            } else if (type == 'simple_k') {
+
+                for (i in seq_len(nLevels-1))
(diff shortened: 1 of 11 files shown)


R/ancova.b.R
```diff
@@ -848,9 +853,16 @@ ancovaClass <- R6::R6Class(
 
             nLevels <- length(levels)
 
-            if (type == 'simple') {
+            if (type %in% c('simple_1', 'simple')) {
```
raviselker: Why is this `simple_1` used as an alias? When using jamovi it still seems to be defined as just `simple` (which I think is also best to keep regarding backwards compatibility).

R/anovarm.b.R
```diff
@@ -894,6 +941,79 @@ anovaRMClass <- R6::R6Class(
         },
 
         #### Helper methods ----
+        .contrastLabels=function(levels, type) {
```
raviselker: This function is almost the same as the one in `an(c)ova` no? Maybe it would be better to move this to the `utilsanova.R` file and then just use the same function in both ANOVAs

R/anovarm.b.R
```diff
@@ -1289,6 +1409,64 @@ calcUnivariateTests <- function(SSP, SSPE, P, df, error_df) {
...
 
+#' Define contrasts
+#' Some contrast are already defined in `emmeans` (simple_1) and just "linked", 
+#' for others (e.g., deviation) the respective function needs to be defined
+#' (cf. https://rdrr.io/cran/emmeans/src/R/emm-contr.R)
+#'
+#' @param type Name / type of contrast (e.g., "simple_1")
+#' @return A function, used as input parameter `method` in emmeans::contrast
+#' @keywords internal
+defineContrasts <- function(type) {
```
raviselker: This function returns a function right? So maybe name it something like `.getContrastFunction()` so that its more clear what it does.

raviselker (changes requested):
Great addition @sjentsch! I have a couple of remarks, but very happy that you implemented this!

R/anovarm.b.R
sjentsch: Done.

R/anovarm.b.R
sjentsch: Done.

R/ancova.b.R
sjentsch: I thought `simple_1` would make it clearer but I agree that it would make things rather more complicated when having an alias. I removed all occurrences of `simple_1`.

sjentsch:
Hei @raviselker, coiuld you have a look at commit [3201c8d](https://github.com/jamovi/jmv/pull/434/commits/3201c8daa8d99cfbc04dd5298ccb4c53d8e73228). The translation markers .() obvisously don't seem to work when moving the contrastLabels outside ancova.b.R and anovarm.b.R. Any suggestions? Thanks and cheers, Sebastian

jonathon-love:
i think you just need to have a reference to `self` in the calling scope. this is how ravi does it:

```
setSingularityWarning = function(self) {
    message <- paste(
        .("Model fit issue detected!"),
        .("It appears that your model has a \"singularity\" problem. This usually means that the model is having trouble distinguishing between certain effects because some variables are too closely related or there isn't enough data to estimate all the effects reliably. To resolve this, consider simplifying the model by removing some of the effects or checking if your data might have redundant or highly correlated variables.")
    )

    setAnalysisNotice(
        self, message, name="refLevelWarning", type=jmvcore::NoticeType$STRONG_WARNING
    )
}
```

https://github.com/jamovi/jmv/blob/master/R/utilsanova.R#L5-L14

whereever you call `.()` ... the `.()` function looks in the calling scope for a `self`

sjentsch:
Hei @raviselker and @jonathon-love, I made the suggested change (using Jonathon's comment from a month ago). I would like to do some cross-checking next week (using Jonathon's SPSS version, I currently don't have access to one). After that, we should be good to go. However, @raviselker you may already have a look at the code in it's current state. If there are changes that you would like me to do, if you make your suggestions before coming Wed, I could implement them and then go on with the cross-checking. Thanks and cheers, Sebastian

jonathon-love:
your mods are including changes to the i18n submodule? i think you haven't quite ironed out your "git hygiene" just yet :P

sjentsch:
Double-checked the results against SPSS. They are identical.

R/ancova.h.R
```diff
@@ -606,8 +607,9 @@ ancovaBase <- if (requireNamespace("jmvcore", quietly=TRUE)) R6::R6Class(
 #' @param qq \code{TRUE} or \code{FALSE} (default), provide a Q-Q plot of
 #'   residuals
 #' @param contrasts a list of lists specifying the factor and type of contrast
-#'   to use, one of \code{'deviation'}, \code{'simple'}, \code{'difference'},
-#'   \code{'helmert'}, \code{'repeated'} or \code{'polynomial'}
+#'   to use, one of \code{'deviation'}, \code{'simple'} (or \code{'simple_1'}),
```
raviselker: `simple_1` is still referenced here. As this is file is auto-generated I think you can just build the module again and this should be fixed.

R/anova.h.R
```diff
@@ -476,8 +477,9 @@ anovaBase <- if (requireNamespace("jmvcore", quietly=TRUE)) R6::R6Class(
 #' @param qq \code{TRUE} or \code{FALSE} (default), provide a Q-Q plot of
 #'   residuals
 #' @param contrasts a list of lists specifying the factor and type of contrast
-#'   to use, one of \code{'deviation'}, \code{'simple'}, \code{'difference'},
-#'   \code{'helmert'}, \code{'repeated'} or \code{'polynomial'}
+#'   to use, one of \code{'deviation'}, \code{'simple'} (or \code{'simple_1'}),
```
raviselker: `simple_1` is still referenced here. As this is file is auto-generated I think you can just build the module again and this should be fixed.

R/anovarm.h.R
```diff
@@ -941,7 +941,10 @@ anovaRMBase <- if (requireNamespace("jmvcore", quietly=TRUE)) R6::R6Class(
 #'   homogeneity of variances (i.e., Levene's test)
 #' @param qq \code{TRUE} or \code{FALSE} (default), provide a Q-Q plot of
 #'   residuals
-#' @param contrasts in development
+#' @param contrasts a list of lists specifying the factor and type of contrast
+#'   to use, one of \code{'deviation'}, \code{'simple'} (or \code{'simple_1'}),
```
raviselker: `simple_1` is still referenced here. As this is file is auto-generated I think you can just build the module again and this should be fixed.

R/utilsanova.R
```diff
@@ -12,3 +12,69 @@ setSingularityWarning = function(self) {
         self, message, name="refLevelWarning", type=jmvcore::NoticeType$STRONG_WARNING
     )
 }
+
+#' Set a warning notice for singularity in the design matrix
```
raviselker: Where does this tile come from? This is not what this function does right?

raviselker (commented):
Some small things, but overall it looks good. As @jonathon-love mentioned, you have indeed also committed the i18n stuff so this needs to be fixed. Maybe jonathon can help you with this.

R/anovarm.h.R
sjentsch: Done, see commits 2726abc5764d4c2fa254cc1c7711a1d9face30a3 and 905cc20bd534abcfb41541b3ad33a289eafa3390.

R/utilsanova.R
sjentsch: Corrected.

R/anova.h.R
sjentsch: Corrected.

R/ancova.h.R
sjentsch: Corrected.

R/ancova.b.R
sjentsch: Removed.

sjentsch:
Hei @raviselker, I removed all instances of simple_1 from the code, and corrected the description for the contrastLabels-function (). I promise to get better with my git hygiene. I had Jonathon's help with cleaning up some of the mess. Sorry for the additional work! Sebastian

raviselker (approved):
Sorry for the delay, but this looks good so let's merge it!

PR #435, opened 2025-09-09
Implemented Data Summary for EFA / PCA

sjentsch:
- shows one table with number of correlations above a minimum (to see if an items is reasonably - a good value would be 0.3 - correlated with the other items) and a maximum value (to see if items are to strongly - a good value would be 0.9 - correlated and should be dropped)
- the respective table is only show if the threshold is set to a value higher than 0 (the default)

### DIFF
--- a/R/efa.h.R
+++ b/R/efa.h.R
@@ -12,6 +12,10 @@ efaOptions <- if (requireNamespace("jmvcore", quietly=TRUE)) R6::R6Class(
             minEigen = 0,
             extraction = "minres",
             rotation = "oblimin",
+            minCorrThr = 0.3,
+            maxCorrThr = 0.9,
+            kmo = FALSE,
+            bartlett = FALSE,
             hideLoadings = 0.3,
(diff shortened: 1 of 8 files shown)


raviselker:
Hi Sebastian,

Thanks for opening the PR! I've had a look and although I think it's useful to be able to have a closer look at the inter-variable correlation, but I'm not sure the data summary is the best way to do this. My main concern is that while it's great for a quick check, it doesn't provide the level of detail you ultimately need for diagnostics. If the table flags an issue (e.g., too many strong correlations), you still have to go find the specific numbers to figure out what's actually going on.

Maybe a better path forward would be to present this information in a more detailed way. I was thinking something similar to how the Residuals for Observed Correlation Matrix is displayed in the CFA module. Here, you get the actual numbers, but you can highlight the ones that are "suspect". That way, users get the full picture without having to look at multiple tables.

What do you think about exploring that approach instead?

Cheers,
Ravi

PR #438, opened 2025-10-25
Update Roxygen, added inst WORDLIST, correct spelling

sjentsch:
One of the requirements of submitting packages to CRAN is that one should have checked for spelling mistakes using `devtools::spell_check()`. This PR will result in that there are no spelling mistakes anymore when running this function. It either corrected the spellings if that were wrong or added words that are correct but not in the dictionary of the spell checker to `inst/WORDLIST`. In addition, Roxygen is upgraded to a more recent version (6.1.1 -> 7.3.3) and some of the mistakes returned by Roxygen were corrected too.


jonathon-love:
merged

PR #441, opened 2025-11-27
Fix failing checks

sjentsch:
These changes should take care of the failing checks on GitHub (oldrel and devel). In addition, some lines of code are copied from PMCMR::posthoc.durbin.test in order to remove the dependency on that package which is deprecated (and will be removed with the next version of R).

The GitHub action for oldrel still fails (with some install.packages issue), however, when trying the code out in a docker / rocker container with R 4.4 (=oldrel) it works without problems.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -25,7 +25,6 @@ Imports:
     car (>= 3.0.0),
     multcomp,
     ggplot2 (>= 2.2.1),
-    PMCMR,
     emmeans (>= 1.4.2),
     vcd,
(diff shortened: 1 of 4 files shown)


jonathon-love:
did you get thorsten to sign off on this? shall we add him to the authors?

sjentsch:
@jonathon-love: I will wrote an e-mail to Thorsten Pohlert (author of PMCMR) with you in CC. Let's wait what he says and then take this further. However, @raviselker can already start to review the first commit.

PR #442, opened 2026-01-14
Feature/implement non integer weights logreg

raviselker:
Moved to manual calculation of AIC, BIC, AUC, r2, and ClassTable as the log likelihood is not calculated correctly for the binomial family in the glm for non-integer weights.

Would be good if someone with some more statistical knowledge could look at this. @sjentsch?


sjentsch:
I would recommend that rather Marcello (@mcfanda) has a look. I strongly assume that he is more knowledgeable about these values (I am primarily an ANOVA guy).

If he can't spare the time, I will have a look.

PR #452, opened 2026-06-10
descriptives: keep integer counts free of decimal places

raviselker:
Frequency counts were typed as 'number' when weights support was added, so even unweighted or integer-weighted counts displayed decimal places. Default the column to integer and switch to number only when the counts are actually non-integer. The case values are unavailable during init, so the type is determined in the run phase.

Closes https://github.com/jamovi/jamovi/issues/1797

### DIFF
--- a/R/descriptives.b.R
+++ b/R/descriptives.b.R
@@ -1000,8 +1000,10 @@ descriptivesClass <- R6::R6Class(
                 # the case values aren't available during init, so set the
                 # counts type here; only show decimals for non-integer counts
                 # (i.e. when weighted by non-integer weights)
-                if ( ! all(freq == round(freq), na.rm=TRUE))
-                    table$addColumn(name='counts', title=.('Counts'), type='number')
+                if ( ! all(freq == round(freq), na.rm=TRUE)) {
+                    countsColumn <- table$columns[['counts']]
+                    countsColumn$.__enclos_env__$private$.type <- 'number'
+                }
 
                 tableVars <- c(var, splitBy)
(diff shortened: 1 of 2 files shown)


R/descriptives.b.R
```diff
@@ -997,6 +997,12 @@ descriptivesClass <- R6::R6Class(
                 table <- tables$get(var)
                 freq <- freqs[[var]]
 
+                # the case values aren't available during init, so set the
+                # counts type here; only show decimals for non-integer counts
+                # (i.e. when weighted by non-integer weights)
+                if ( ! all(freq == round(freq), na.rm=TRUE))
+                    table$addColumn(name='counts', title=.('Counts'), type='number')
```
jonathon-love: this is relying on some undocumented feature, where if you add a column with the same name as an existing column, it replaces it?

this would be a different way to do it:

`table$columns[['counts']]$.__enclos_env__$private$.type <- 'number'`

raviselker: Went with the `private$.type` approach. One caveat: it can't be a single chained assignment —

    table$columns[['counts']]$.__enclos_env__$private$.type <- 'number'

R desugars that into get-then-set-back calls up to the top-level target, so the final step writes the column list back through `table$columns`, which is a read-only active binding. That throws `unused argument (... list(<environment> x4))`. Extracting the column into a local first avoids the writeback:

    countsColumn <- table$columns[['counts']]
    countsColumn$.__enclos_env__$private$.type <- 'number'

